PR #4, opened 2020-02-25
Reduce Boost dependencies; switch to V8 8.0

tudorms:
* Switch back to V8 8.0 (with disabled FH4 support for 141 toolset compatibility);
* Reduce dependencies on Boost libraries;
* Commit a local clone of Boost ASIO to significantly improve CI loop execution time;
* Misc fixes (include v8 version / ref number in NuGet, etc.);


JunielKatarn:
Why copy sources instead of using a submodule or NuGet package?
This makes the dependency harder to track or upgrade.

tudorms:
> Why copy sources instead of using a submodule or NuGet package?
> This makes the dependency harder to track or upgrade.

There is unfortunately no way to submodule a library + all its dependencies transitively for Boost. Is there a NuGet that includes just the ASIO library + its dependencies? We'll probably end up in the same situation we have today where we need to get the whole of Boost as a NuGet (or a large subset of it) so it wouldn't be much faster than downloading the .zip / .7z.

JunielKatarn:
> There is unfortunately no way to submodule a library + all its dependencies transitively for Boost. Is there a NuGet that includes just the ASIO library + its dependencies? We'll probably end up in the same situation we have today where we need to get the whole of Boost as a NuGet (or a large subset of it) so it wouldn't be much faster than downloading the .zip / .7z.

Yeah, submodules would be hard to manage if there are transitive dependencies.
The source-only NuGet available at nuget.org is the easiest option, which also avoids doing manual copies of source files.

I strongly suggest to take that approach, which also makes updates trivial.

tudorms:
> > There is unfortunately no way to submodule a library + all its dependencies transitively for Boost. Is there a NuGet that includes just the ASIO library + its dependencies? We'll probably end up in the same situation we have today where we need to get the whole of Boost as a NuGet (or a large subset of it) so it wouldn't be much faster than downloading the .zip / .7z.
> 
> Yeah, submodules would be hard to manage if there are transitive dependencies.
> The source-only NuGet available at nuget.org is the easiest option, which also avoids doing manual copies of source files.
> 
> I strongly suggest to take that approach, which also makes updates trivial.

The problem this is trying to solve is that downloading a zip archive with the boost source code and unzipping it takes a lot of time; I believe the Boost NuGet is the same thing, a zip archive with the source code, unless I'm missing something.

JunielKatarn:
> > > There is unfortunately no way to submodule a library + all its dependencies transitively for Boost. Is there a NuGet that includes just the ASIO library + its dependencies? We'll probably end up in the same situation we have today where we need to get the whole of Boost as a NuGet (or a large subset of it) so it wouldn't be much faster than downloading the .zip / .7z.
> > 
> > 
> > Yeah, submodules would be hard to manage if there are transitive dependencies.
> > The source-only NuGet available at nuget.org is the easiest option, which also avoids doing manual copies of source files.
> > I strongly suggest to take that approach, which also makes updates trivial.
> 
> The problem this is trying to solve is that downloading a zip archive with the boost source code and unzipping it takes a lot of time; I believe the Boost NuGet is the same thing, a zip archive with the source code, unless I'm missing something.

Here is a sample of a worst case scenario (NuGet not cached, so it needs to be downloaded):
```
PS D:\erasme> Measure-Command { nuget install -OutputDirectory D:\erasme\ boost -Version 1.71.0 }                       

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 22
Milliseconds      : 319
Ticks             : 223196260
TotalDays         : 0.00025832900462963
TotalHours        : 0.00619989611111111
TotalMinutes      : 0.371993766666667
TotalSeconds      : 22.319626
TotalMilliseconds : 22319.626
```

With the NuGet in the machine's cache, it would take even less.
Subjectively, I don't believe this step would be negatively impactful to productivity, and prevents us from having to track files manually.

tudorms:
> > > > There is unfortunately no way to submodule a library + all its dependencies transitively for Boost. Is there a NuGet that includes just the ASIO library + its dependencies? We'll probably end up in the same situation we have today where we need to get the whole of Boost as a NuGet (or a large subset of it) so it wouldn't be much faster than downloading the .zip / .7z.
> > > 
> > > 
> > > Yeah, submodules would be hard to manage if there are transitive dependencies.
> > > The source-only NuGet available at nuget.org is the easiest option, which also avoids doing manual copies of source files.
> > > I strongly suggest to take that approach, which also makes updates trivial.
> > 
> > 
> > The problem this is trying to solve is that downloading a zip archive with the boost source code and unzipping it takes a lot of time; I believe the Boost NuGet is the same thing, a zip archive with the source code, unless I'm missing something.
> 
> Here is a sample of a worst case scenario (NuGet not cached, so it needs to be downloaded):
> 
> ```
> PS D:\erasme> Measure-Command { nuget install -OutputDirectory D:\erasme\ boost -Version 1.71.0 }                       
> 
> Days              : 0
> Hours             : 0
> Minutes           : 0
> Seconds           : 22
> Milliseconds      : 319
> Ticks             : 223196260
> TotalDays         : 0.00025832900462963
> TotalHours        : 0.00619989611111111
> TotalMinutes      : 0.371993766666667
> TotalSeconds      : 22.319626
> TotalMilliseconds : 22319.626
> ```
> 
> With the NuGet in the machine's cache, it would take even less.
> Subjectively, I don't believe this step would be negatively impactful to productivity, and prevents us from having to track files manually.

Thank you, let me take a look

PR #5, opened 2020-02-25
Switch to V8 8.0; improve CI build time

tudorms:
* Switch back to V8 8.0 (with disabled FH4 support for 141 toolset compatibility);
* Reduce unnecessary dependencies on Boost libraries;
* Use NuGet for bringing in Boost ASIO headers (much faster than downloading the whole zip from boostorg);
* Misc fixes (include v8 version / ref number in NuGet, etc.);


.gitignore
```diff
@@ -1,2 +1,3 @@
 build
-out
\ No newline at end of file
+out
+.vscode
```
NickGerleman: tubonit: leaving a blank line at the end of the file helps with diffs (and will avoid the red warning Github gives you)

src/inspector/inspector_agent.cpp
```diff
@@ -42,9 +40,21 @@ std::string GetProcessTitle() {
 }
 
 std::string GenerateID() {
-  boost::uuids::uuid uuid = boost::uuids::random_generator()();
-  return boost::uuids::to_string(uuid);
-  //return "RANDON_UUID";
+  static std::random_device rd;
+  static std::mt19937 mte(rd());
+
+  std::uniform_int_distribution<int> dist(1, 60000);
```
NickGerleman: Why 60000? The default is max int if unspeified.

NickGerleman: Similarly, should we use uint16_t as a template parameter if that's what we're filling?

PR #7, opened 2020-03-03
Delay load v8jsi.dll

tudorms:
* delay load the dll by default (so that consumers can have a run-time / optional dependency on v8jsi.dll without needing to distribute it if unused);
* the Platform must be process-singleton to match v8 internals (V8 is not embedder-friendly and makes several assumptions about objects that are initialized once per process - we're matching that design with our V8Platform to avoid crashes if multiple contexts are created / shut down during a process' lifetime);

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -1128,7 +1128,30 @@ bool V8Runtime::isFunction(const jsi::Object &obj) const {
 
 bool V8Runtime::isHostObject(const jsi::Object &obj) const {
   _ISOLATE_CONTEXT_ENTER
-  std::abort();
+  if (objectRef(obj)->InternalFieldCount() < 1) {
+    return false;
+  }
+
+  auto internalFieldRef = objectRef(obj)->GetInternalField(0);
+  if (internalFieldRef.IsEmpty()) {
+    return false;
+  }
+
+  v8::Local<v8::External> internalField = v8::Local<v8::External>::Cast(internalFieldRef);
+  if (!internalField->Value()) {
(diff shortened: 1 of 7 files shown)


src/V8JsiRuntime.cpp
```diff
@@ -1128,7 +1128,30 @@ bool V8Runtime::isFunction(const jsi::Object &obj) const {
...
+  }
+
+  v8::Local<v8::External> internalField = v8::Local<v8::External>::Cast(internalFieldRef);
+  if (!internalField->Value()) {
+    return false;
+  }
+
+  HostObjectProxy *hostObjectProxy = reinterpret_cast<HostObjectProxy *>(internalField->Value());
+
+  for (std::shared_ptr<HostObjectLifetimeTracker> hostObjectLifetimeTracker :
```
NickGerleman: nit: this should be by reference instead of value

src/V8JsiRuntime_impl.h
```diff
@@ -114,8 +116,10 @@ while (!use_count_s_.compare_exchange_weak(current, current - 1))
   ;*/
 
     if (--use_count_s_ == 0) {
-      v8::V8::ShutdownPlatform();
-      platform_s_ = nullptr;
+      // We cannot shutdown the platform once created because V8 internally references bits of the platform from process-globals
```
NickGerleman: Do we need to change the API for this if it never dies? Could we track this somewhere?

tudorms: The public API didn't change, but overall this could use some cleanup.

PR #8, opened 2020-03-19
Convert JS errors into C++ jsi::JSError exceptions; update JSI headers

tudorms:
- convert JS errors into C++ jsi::JSError exceptions to match Hermes (for now this covers only function calls);
- update JSI headers to latest available (API compatible, this includes an UTF8 conversion issue fix we found in Office LPC);
- remove some console logging that was leftover from previous implementations;

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -368,9 +368,6 @@ struct SameCodeObjects {
     const v8::JitCodeEvent *event) {
   switch (event->type) {
     case v8::JitCodeEvent::CODE_ADDED:
-      if (event->code_type == v8::JitCodeEvent::CodeType::BYTE_CODE) {
-        std::cout << "B!!";
-      }
 #ifdef _WIN32
       EventWriteJIT_CODE_EVENT(
           event->type,
@@ -1245,11 +1242,16 @@ jsi::Value V8Runtime::call(
     argv.push_back(valueRef(args[i]));
   }
 
+  v8::TryCatch trycatch(isolate_);
   v8::MaybeLocal<v8::Value> result = func->Call(
       isolate_->GetCurrentContext(),
       valueRef(jsThis),
       static_cast<int>(count),
       argv.data());
+
+  if (trycatch.HasCaught()) {
+    ReportException(&trycatch);
+  }
 
   // Call can return
   if (result.IsEmpty()) {
@@ -1271,6 +1273,7 @@ jsi::Value V8Runtime::callAsConstructor(
     argv.push_back(valueRef(args[i]));
   }
 
+  v8::TryCatch trycatch(isolate_);
   v8::Local<v8::Object> newObject;
   if (!func->NewInstance(
                GetIsolate()->GetCurrentContext(),
@@ -1278,6 +1281,10 @@ jsi::Value V8Runtime::callAsConstructor(
                argv.data())
            .ToLocal(&newObject)) {
     new jsi::JSError(*this, "Object construction failed!!");
+  }
+
+  if (trycatch.HasCaught()) {
+    ReportException(&trycatch);
   }
 
   return createValue(newObject);
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -364,8 +364,7 @@ struct SameCodeObjects {
   }
 };
 
-/*static */ void V8Runtime::JitCodeEventListener(
-    const v8::JitCodeEvent *event) {
+/*static */ void V8Runtime::JitCodeEventListener(const v8::JitCodeEvent *event) {
   switch (event->type) {
     case v8::JitCodeEvent::CODE_ADDED:
 #ifdef _WIN32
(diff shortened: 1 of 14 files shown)


src/jsi/decorator.h
```diff
@@ -335,18 +335,18 @@ class RuntimeDecorator : public Base, private jsi::Instrumentation {
     plain().instrumentation().collectGarbage();
   }
 
-  bool createSnapshotToFile(const std::string& path) override {
-    return plain().instrumentation().createSnapshotToFile(path);
+  void createSnapshotToFile(const std::string& path) override {
```
hansenyy: >void [](start = 2, length = 4)

These seems to be the same as the latest JSI files from Hermes/RN master. Could you please confirm this?

src/V8JsiRuntime.cpp
```diff
@@ -368,9 +368,6 @@ struct SameCodeObjects {
     const v8::JitCodeEvent *event) {
   switch (event->type) {
```
NikoAri: >  s [](start = 0, length = 3)

(nit) can you fix spacing?

src/V8JsiRuntime.cpp
```diff
@@ -1245,12 +1242,17 @@ jsi::Value V8Runtime::call(
     argv.push_back(valueRef(args[i]));
   }
 
+  v8::TryCatch trycatch(isolate_);
```
NikoAri: >trycatch [](start = 15, length = 8)

(nit) would it make sense to add scope for this variable?

src/jsi/jsi-inl.h
```diff
@@ -30,7 +30,7 @@ inline Value toValue(Runtime& runtime, const char* str) {
   return String::createFromAscii(runtime, str);
 }
 inline Value toValue(Runtime& runtime, const std::string& str) {
-  return String::createFromAscii(runtime, str);
+  return String::createFromUtf8(runtime, str);
```
NikoAri: >createFromUtf8 [](start = 17, length = 14)

Thanks!

src/jsi/jsi.cpp
```diff
@@ -37,6 +37,27 @@ std::string kindToString(const Value& v, Runtime* rt = nullptr) {
...
+        std::string("callGlobalFunction: JS global property '") + name +
+        "' is " + kindToString(v, &runtime) + ", expected a Function");
+  }
+  Object o = v.getObject(runtime);
+  if (!o.isFunction(runtime)) {
+    throw JSINativeException(
+        std::string("callGlobalFunction: JS global property '") + name +
+        "' is a non-callable Object, expected a Function");
+  }
+  Function f = std::move(o).getFunction(runtime);
```
NikoAri: >std::move( [](start = 15, length = 10)

Why do we need std::move here? Am I reading parenthesis correctly?

src/jsi/jsi.cpp
```diff
@@ -347,7 +368,11 @@ JSError::JSError(Runtime& rt, Value&& value) {
 JSError::JSError(Runtime& rt, std::string msg) : message_(std::move(msg)) {
   try {
     setValue(
-        rt, rt.global().getPropertyAsFunction(rt, "Error").call(rt, message_));
+        rt,
+        callGlobalFunction(rt, "Error", String::createFromUtf8(rt, message_)));
+  } catch (const std::exception& ex) {
```
NikoAri: >exception [](start = 22, length = 9)

Should we catch a more specific exception here?

src/jsi/jsi.cpp
```diff
@@ -347,7 +368,11 @@ JSError::JSError(Runtime& rt, Value&& value) {
...
+    message_ = std::string(ex.what()) + " (while raising " + message_ + ")";
+    setValue(rt, String::createFromUtf8(rt, message_));
```
NikoAri: >setValue [](start = 4, length = 8)

What if this one throws? Or we only want to catch "callGlobalFunction"? If so should we wrap it to "callGlobalFunction" only?

src/jsi/jsi.cpp
```diff
@@ -360,6 +385,8 @@ JSError::JSError(Runtime& rt, std::string msg, std::string stack)
     e.setProperty(rt, "message", String::createFromUtf8(rt, message_));
     e.setProperty(rt, "stack", String::createFromUtf8(rt, stack_));
     setValue(rt, std::move(e));
+  } catch (const std::exception& ex) {
```
NikoAri: >catch [](start = 4, length = 5)

on the overall design. Why are we moving toward exception based model? This pattern feels very tricky thus could be hard to get right. Would it make sense to go with model of something like:

bool Foo(args, Error& errorMessage) noexcept {...}

src/jsi/jsi-inl.h
tudorms: Note that all files under src/jsi are simply copies from Facebook's Hermes project. They need to stay in sync with the upstream repo so I won't make any changes there.

src/jsi/decorator.h
tudorms: Yes, correct, these are the latest headers from hermes/API/jsi

PR #15, opened 2020-06-05
Fix basic debugging support. Upgrade to V8 8.4

tudorms:
- Fix a couple of bugs around the inspector (crash on detach, hang on shutdown with debugger attached, etc.);
- Upgrade to V8 version 8.4;

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/15)

### DIFF
--- a/src/inspector/inspector_agent.cpp
+++ b/src/inspector/inspector_agent.cpp
@@ -20,6 +20,8 @@
 #include <random>
 #include <algorithm>
 
+#include "../V8Platform.h"
+
 namespace inspector {
 
 const char TAG_CONNECT[] = "#connect";
@@ -411,6 +413,8 @@ void AgentImpl::Start() {
         std::unique_ptr<InspectorAgentDelegate>(&delegate), port_);
     server_ = &server;
 
+    state_ = State::kAccepting;
+
     // This loops
     if (!server.Start()) {
       std::abort();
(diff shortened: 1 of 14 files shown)


src/inspector/inspector_agent.cpp
```diff
@@ -531,7 +535,16 @@ void AgentImpl::PostIncomingMessage(
   if (AppendMessage(
           &incoming_message_queue_, session_id, Utf8ToStringView(message))) {
 
-    platform_.GetForegroundTaskRunner(isolate_)->PostTask(std::make_unique<DispatchOnInspectorBackendTask>(this));
+  std::shared_ptr<v8::TaskRunner> foregroundTaskRunner;
+
+#ifdef USE_DEFAULT_PLATFORM
```
mganandraj, line 540: Tudor.. do inspector currently work with default platform now ? 

Last i tested, there was some performance issue with our platform implementation.. Essentially everything was running slower.. and it started happening after we fixed an issue in LPC with the Date() constructor.. the fix was in the couple of time related functions in our platform implementation ..

tudorms: No, our platform doesn't currently work, we're building with USE_DEFAULT_PLATFORM; but wanted to keep that code path alive as well just in case. With our platform, no async code runs (I presume because we're missing a message pumping step).

---
In reply to: [436070132](https://github.com/microsoft/v8-jsi/pull/15#discussion_r436070132) [](ancestors = 436070132)

PR #17, opened 2020-06-30
Some inspector code cleanup

tudorms:
Add safety to inspector code (by replacing most naked pointer usage with smart pointers). Simplify the structure by getting rid of unnecessary wrappers where possible.
**Note**: the inspector is more stable than before but still not quite "ship quality" ready.

(Experimentally) add support for platform shutdown by keeping track of usages. This needs more testing as there are several places in V8 where "process globals / statics" are keeping raw pointers to the platform's internals.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/17)

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -83,13 +83,22 @@ class TaskRunnerAdapter : public v8::TaskRunner {
         std::make_unique<TaskAdapter>(std::move(task)), delay_in_seconds);
   }
 
+  bool IdleTasksEnabled() override {
+    return taskRunner_->IdleTasksEnabled();
+  }
+
   void PostIdleTask(std::unique_ptr<v8::IdleTask> task) override {
     taskRunner_->postIdleTask(
         std::make_unique<IdleTaskAdapter>(std::move(task)));
   }
 
-  bool IdleTasksEnabled() override {
-    return taskRunner_->IdleTasksEnabled();
+  void PostNonNestableTask(std::unique_ptr<v8::Task> task) override {
+    //TODO: non-nestable
+    taskRunner_->postTask(std::make_unique<TaskAdapter>(std::move(task)));
+  }
+
+  bool NonNestableTasksEnabled() const override {
+    return true;
   }
 
  private:
@@ -456,7 +465,7 @@ v8::Isolate *V8Runtime::CreateNewIsolate() {
       std::move(args_.foreground_task_runner));
   isolate_->SetData(
       v8runtime::ISOLATE_DATA_SLOT,
-      new v8runtime::IsolateData({&foreground_task_runner_, this}));
+      new v8runtime::IsolateData({foreground_task_runner_}));
 
   v8::Isolate::Initialize(isolate_, create_params_);
 
@@ -572,6 +581,12 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
 }
 
 V8Runtime::~V8Runtime() {
+#ifdef _WIN32
+  if (inspector_agent_ && inspector_agent_->IsStarted()) {
+    inspector_agent_->stop();
+  }
+#endif
+
   host_object_constructor_.Reset();
   context_.Reset();
 
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -552,6 +552,7 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
   if (tls_isolate_usage_counter_++ > 0) {
     isolate_ = v8::Isolate::GetCurrent();
   } else {
+    platform_holder_.addUsage();
     CreateNewIsolate();
   }
 
@@ -581,10 +582,12 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
 }
 
 V8Runtime::~V8Runtime() {
+  // TODO: add check that destruction happens on the same thread id as construction
 #ifdef _WIN32
   if (inspector_agent_ && inspector_agent_->IsStarted()) {
     inspector_agent_->stop();
   }
+  inspector_agent_.reset();
 #endif
 
   host_object_constructor_.Reset();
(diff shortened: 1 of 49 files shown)


src/V8JsiRuntime_impl.h
```diff
@@ -90,10 +90,11 @@ class V8PlatformHolder {
   }
 
   V8PlatformHolder() {
+    //addUsage();
```
NikoAri, line 93: >addUsage [](start = 6, length = 8)

remove?

src/inspector/inspector_agent.cpp
```diff
@@ -405,20 +374,23 @@ void InspectorWrapConsoleCall(const v8::FunctionCallbackInfo<v8::Value> &args) {
 }
 
 void AgentImpl::Start() {
-  std::thread([this]() {
-    InspectorAgentDelegate delegate(this, "", script_name_, wait_);
-    delegate_ = &delegate;
+  auto self(shared_from_this());
+  std::thread([this, self]() {
```
NikoAri, line 378: >this, self] [](start = 15, length = 11)

(nit) Is that a good pattern to pass both "this" and "self"? why not just pass "self" and use it in the function?

tudorms: It is pretty strange, and needs to be followed up on, but I copied existing code in the inspector that does the same thing.

---
In reply to: [447994164](https://github.com/microsoft/v8-jsi/pull/17#discussion_r447994164) [](ancestors = 447994164)

src/inspector/inspector_socket.cpp
```diff
@@ -453,7 +435,7 @@ class HttpEvent {
 
 class HttpHandler : public ProtocolHandler {
  public:
-  explicit HttpHandler(InspectorSocket* inspector, TcpHolder::Pointer tcp)
+  explicit HttpHandler(InspectorSocket* inspector, std::shared_ptr<TcpHolder> tcp)
```
NikoAri, line 438: >> [](start = 76, length = 1)

(nit) this could be &&, which is common for constuctors

NikoAri: same on line 602


---
In reply to: [447994884](https://github.com/microsoft/v8-jsi/pull/17#discussion_r447994884) [](ancestors = 447994884)

src/inspector/inspector_socket.h
```diff
@@ -5,25 +5,20 @@
 
 #include <string>
 #include <vector>
+#include <memory>
```
NikoAri, line 8: ><m [](start = 9, length = 2)

(nit) alphabetize

src/inspector/inspector_socket.h
```diff
@@ -35,24 +30,16 @@ class InspectorSocket {
     virtual ~Delegate() {}
   };
 
-  using DelegatePointer = std::unique_ptr<Delegate>;
-  using Pointer = std::unique_ptr<InspectorSocket>;
-
-  static Pointer Accept(std::shared_ptr<tcp_connection> connection, DelegatePointer delegate);
-
-  ~InspectorSocket();
+  static std::unique_ptr<InspectorSocket> Accept(std::shared_ptr<tcp_connection> connection, std::unique_ptr<Delegate> delegate);
```
NikoAri, line 33: >> [](start = 117, length = 1)

Shouldn't unique_ptr be mostly &&?

src/inspector/inspector_socket_server.h
```diff
@@ -16,24 +16,39 @@ class InspectorSocketServer;
...
-  virtual void MessageReceived(int session_id, const std::string& message) = 0;
-  virtual std::vector<std::string> GetTargetIds() = 0;
-  virtual std::string GetTargetTitle(const std::string& id) = 0;
-  virtual std::string GetTargetUrl(const std::string& id) = 0;
-  virtual ~SocketServerDelegate() {}
+class InspectorAgentDelegate {
+ public:
+  InspectorAgentDelegate(
+      AgentImpl& agent,
+      const std::string &script_path,
```
NikoAri, line 23: > & [](start = 23, length = 2)

(nit) could these be made && to avoid copying?

src/inspector/inspector_tcp.cpp
```diff
@@ -30,25 +25,29 @@ void tcp_server::run() {
...
+
+  io_service_.stop();
+}
+
 void tcp_server::do_accept()
 {
+  std::shared_ptr<tcp_server> self;
   acceptor_.async_accept(socket_,
-    [this](boost::system::error_code ec)
+    [this, self](boost::system::error_code ec)
```
NikoAri, line 40: >[this, self [](start = 4, length = 11)

why is this lifetime needed here?

src/V8JsiRuntime.cpp
```diff
@@ -581,10 +582,12 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
 }
 
 V8Runtime::~V8Runtime() {
+  // TODO: add check that destruction happens on the same thread id as construction
```
mganandraj, line 585: >// TODO: add check that destruction happens on the same thread id as construction [](start = 2, length = 81)

This is what RN/Hermes does : [https://github.com/facebook/react-native/blob/42c1957aff726b7f0a29eb26e81b69bb50bc4710/ReactAndroid/src/main/java/com/facebook/hermes/reactexecutor/HermesExecutorFactory.cpp#L80](https://github.com/facebook/react-native/blob/42c1957aff726b7f0a29eb26e81b69bb50bc4710/ReactAndroid/src/main/java/com/facebook/hermes/reactexecutor/HermesExecutorFactory.cpp#L80)

src/inspector/inspector_tcp.cpp
mganandraj: As far as i remember .. the shared pointer capture is required to keep the object alive .. This is a model followed by boost::asio ... "this" capture was added as a shortcut to be able to access the private variables..

---
In reply to: [447997236](https://github.com/microsoft/v8-jsi/pull/17#discussion_r447997236) [](ancestors = 447997236)

src/inspector/inspector_socket_server.h
tudorms: It looks like the copies are needed due to threading and lifetimes.

---
In reply to: [447996815](https://github.com/microsoft/v8-jsi/pull/17#discussion_r447996815) [](ancestors = 447996815)

PR #20, opened 2020-08-10
Add UWP (and ARM64) targets

tudorms:
- Update V8 to version 8.5
- Add targets for UWP for x86, x64 and ARM64 (**note**: we're not signing up to support UWP moving forward, this could be a one-time drop if future V8 updates break our UWP patches).

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/20)


ReactNative.V8Jsi.Windows.targets
```diff
@@ -1,31 +1,24 @@
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Set V8_ARCH to "win32" for desktop builds -->
+    <V8_ARCH Condition="'$(V8_ARCH)' == ''">uwp</V8_ARCH>
```
JunielKatarn, line 4: (nit)

The convention for MSBuild properties is Pascal Case.

Suggestion: Call those `V8Arch`, `V8Platform`, etc.

NickGerleman:
Out of curiosity is the UWP target interpreter only? I was under the impression we couldn't JIT on UWP since VirtualProtect is banned (ignoring cleverness to bypass WAC).

tudorms:
Last time I tested WACK (months ago), it passed (with default JIT enabled).

---
In reply to: [671646713](https://github.com/microsoft/v8-jsi/pull/20#issuecomment-671646713) [](ancestors = 671646713)

NickGerleman:
Oh interesting. I wonder how that works internally.

localbuild.ps1
```diff
@@ -29,17 +27,32 @@ if ($Setup) {
     }
 }
 
-if ($Platform -like "all") {
-    foreach ($Plat in $Platforms) {
-        foreach ($Config in $Configurations) {
-            Write-Host "Building $Plat $Config..."
-            & ".\scripts\build.ps1" -SourcesPath $SourcesPath -OutputPath $OutputPath -Platform $Plat -Configuration $Config
+if ($Configuration -eq "all") {
```
JunielKatarn, line 30: Suggestion: Use `ValidateSet`, and allow an explicit list of architectures as opposed to "special" metavalues.

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7#attributes-of-parameters

This also removes arg-validated boilerplate code.

scripts/build.ps1
```diff
@@ -83,21 +86,21 @@ if (!(Test-Path -Path "$OutputPath\build\native\include\jsi")) {
...
+if (!(Test-Path -Path "$OutputPath\lib\$Architecture\$Configuration\$Platform")) {
+    New-Item -ItemType "directory" -Path "$OutputPath\lib\$Architecture\$Configuration\$Platform" | Out-Null
 }
 
 # Binaries
 if (!$PSVersionTable.Platform -or $IsWindows) {
-    Copy-Item "$buildoutput\v8jsi.dll" -Destination "$OutputPath\lib\$Configuration\$Platform"
-    Copy-Item "$buildoutput\v8jsi.dll.lib" -Destination "$OutputPath\lib\$Configuration\$Platform"
-    Copy-Item "$buildoutput\v8jsi.dll.pdb" -Destination "$OutputPath\lib\$Configuration\$Platform"
+    Copy-Item "$buildoutput\v8jsi.dll" -Destination "$OutputPath\lib\$Architecture\$Configuration\$Platform"
```
JunielKatarn, line 95: Out of curiosity, shouldn't MSBuild naturally take care of this as longa s `IntDir` and `OutDir` are correctly set? (See RNW's `Directory.Build.props`).

tudorms: V8 doesn't use MSBuild, it uses proprietary build tools built on top of GN

---
In reply to: [468251030](https://github.com/microsoft/v8-jsi/pull/20#discussion_r468251030) [](ancestors = 468251030)

localbuild.ps1
tudorms: Cool, I'm still a PowerShell newb :)

---
In reply to: [468250255](https://github.com/microsoft/v8-jsi/pull/20#discussion_r468250255) [](ancestors = 468250255)

JunielKatarn: Aren't we all? :)

Still, if there's hurry to merge this, we can always make incremental upgrades. The script seems correct to me.

PR #21, opened 2020-08-11
Publish symbols in separate package

tudorms:
With UWP and WIN32 in the same NuGet, the resulting nupkg size is ~540Mb, whereas nuget.org has a 250Mb package size limit.

There is support for snupkg (symbol packages) in [nuget](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg), but very little documentation on the Azure DevOps side.

This removed the PDBs from the main v8jsi NuGet, we're still exploring options for where to upload the symbols.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/21)

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -66,7 +66,10 @@ if (!$?) {
     exit 1
 }
 
-& ninja -v -j 16 -C $buildoutput v8jsi jsitests | Tee-Object -FilePath "$SourcesPath\build.log"
+# We'll use 2x the number of cores for parallel execution
+$numberOfThreads = [int]((Get-CimInstance Win32_ComputerSystem).NumberOfLogicalProcessors) * 2
+
+& ninja -v -j $numberOfThreads -C $buildoutput v8jsi jsitests | Tee-Object -FilePath "$SourcesPath\build.log"
 if (!$?) {
     Write-Host "Build failure, check logs for details"
     exit 1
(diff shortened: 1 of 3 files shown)


JunielKatarn:
Note, nuget.org does not yet support native SNUPKG files. Meaning, the symbol package won't be correctly uploaded.
https://devblogs.microsoft.com/nuget/improved-package-debugging-experience-with-the-nuget-org-symbol-server/

It's good that we produce the separate package, but we may need to host them elsewhere (@dannyvv, maybe you can comment on how to achieve this?).

scripts/build.ps1
```diff
@@ -66,7 +66,10 @@ if (!$?) {
     exit 1
 }
 
-& ninja -v -j 16 -C $buildoutput v8jsi jsitests | Tee-Object -FilePath "$SourcesPath\build.log"
+# We'll use 2x the number of cores for parallel execution
```
NikoAri, line 69: >2x [](start = 12, length = 2)

why? we shouldn't be meddling with such low level system stuff. can't system figure out most optimal parallelism for us?

tudorms: With a hardcoded 16 parallelism, some build agents were choking (effectively grinding to a halt). This tries to reduce the number of parallel builds on low-end devices like that.

PR #24, opened 2020-08-20
Create versioning scheme for RN compatibility

tudorms:
We'll have version 0.63.* for ReactNative 0.60 - 0.63 and 0.64.* for ReactNative 0.64+

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/24)


NickGerleman:
@tudorms if you're updating JSI in the same PR, would you be able to make the change to stop packing JSI headers into the NuGet package, or scope them into a place where there aren't conflicts? We will otherwise still get forced header redirection by including the package.

PR #30, opened 2020-10-09
Update to version 8.6

tudorms:
A few more breaking changes this time around, this tries to upgrade to V8 version 8.6 and fix the windows builds.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/30)

### DIFF
--- a/scripts/patch/build.diff
+++ b/scripts/patch/build.diff
@@ -1,8 +1,8 @@
 diff --git a/config/compiler/BUILD.gn b/config/compiler/BUILD.gn
-index 01d97a360..6ee2813ef 100644
+index 06c166f71..e047333af 100644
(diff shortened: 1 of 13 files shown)


scripts/patch/build.diff
```diff
@@ -1,8 +1,8 @@
 diff --git a/config/compiler/BUILD.gn b/config/compiler/BUILD.gn
```
JunielKatarn, line 1: Out of curiosity, how are we generating these patches?

tudorms: Manually :(

If we clean up some of this, perhaps we can push a PR upstream to Google.

PR #33, opened 2020-10-30
Further simplify CI workflow

JunielKatarn:
- Removes deprecated pull request YML files.
- Remove NUSPEC and TARGETS duplcation.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/33)


.ado/pull-request.yml
```diff
@@ -99,8 +99,10 @@ jobs:
       matrix:
         Desktop:
           NuGetId: ReactNative.V8Jsi.Windows
+          PdbExtension: ipdb
```
tudorms, line 102: >PdbExtension [](start = 10, length = 12)

[nit] Maybe we can call this "file extensions to ignore" or something like that. The goal was actually to not include PDBs in the UWP build because we're trying to keep its size below the 250Mb limit (which may no longer be needed if we separate the symbols or use the ADO feed instead).

tudorms (approved):
:shipit:

.ado/pull-request.yml
JunielKatarn: You got it. I'll name it `NuGetExcludePattern`.

PR #35, opened 2020-11-11
Fix arm64 build

tudorms:
Previous commit (update to 8.7) broke the ARM64 build. This cherry-picks [the fix](https://chromium-review.googlesource.com/c/v8/v8/+/2440717) from upstream V8 as a patch.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/v8-jsi/pull/35)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.4",
-    "v8ref": "refs/branch-heads/8.6"
+    "version": "0.64.5",
+    "v8ref": "refs/branch-heads/8.7"
 }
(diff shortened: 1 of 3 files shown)


JunielKatarn (approved):
What is exactly the fix?

Omitting unwanted changes from upstream?
Shrinking the local patching in BUILD.gn?

tudorms:
I should've created a new branch for better history; but basically the fix is in iteration 2: a change in src/compiler/c-linkage.cc

---
In reply to: [528648454](https://github.com/microsoft/v8-jsi/pull/35#issuecomment-528648454) [](ancestors = 528648454)

PR #42, opened 2021-02-04
Replace custom patch with cherry-pick from V8 9.0 / master

tudorms:
Replace the custom patch we had for the MSVC debug break on vector custom allocator with cherry-picking the fix in upstream (until we get 9.0): https://chromium-review.googlesource.com/c/v8/v8/+/2622915

V8 8.9 was officially released today, this will rebuild the package with the new commit id from branch head (255.16): https://v8.dev/blog/v8-release-89

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/42)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.8",
+    "version": "0.64.9",
     "v8ref": "refs/branch-heads/8.9"
 }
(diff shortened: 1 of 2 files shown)


NikoAri (approved):
:shipit:

PR #44, opened 2021-03-01
More tracing and instrumentation, Inspector fixes, and enabling them by default

mganandraj:
This PR has the following changes,
1. More trace and instrumentation in runtime and inspector for debugging and monitoring using ETW, and a script for real time trace consumption
2. Inspector cleanups and fixes. We have client validation, automatic session discovery from chrome inspector, CPU profiling, target validation etc. Our inspector is now at part with NodeJS inspector.
3. We are now enabling inspector and instrumentation by default. In the coming days we will closely observe the impacts of keeping them turned on. Also, will work on tools over the generated traces.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/44)


src/inspector/inspector_agent.cpp
```diff
@@ -374,26 +379,20 @@ void InspectorWrapConsoleCall(const v8::FunctionCallbackInfo<v8::Value> &args) {
...
-
-    state_ = State::kAccepting;
-
-    // This loops
-    if (!server_->Start()) {
-      std::abort();
-    }
-
-    server_->Stop();
+  // auto self(shared_from_this());
```
NikoAri, line 382: >// a [](start = 2, length = 4)

Remove

src/inspector/inspector_socket.cpp
```diff
@@ -339,17 +345,24 @@ class WsHandler : public ProtocolHandler {
...
-  void CancelHandshake() override {}
+  void AcceptUpgrade(const std::string& accept_key) override {
+    TRACEV8INSPECTOR_VERBOSE("WsHandler::AcceptUpgrade");
+  }
+  void CancelHandshake() override {
+    TRACEV8INSPECTOR_VERBOSE("WsHandler::CancelHandshake");
+  }
 
   void OnEof() override {
+    TRACEV8INSPECTOR_VERBOSE("WsHandler::OnEof");
```
NikoAri, line 356: >TRACEV8INSPECTOR_VERBOSE [](start = 4, length = 24)

For my understanding, is there a way we can collect those during Watson crashes?

src/inspector/inspector_tcp.cpp
```diff
@@ -88,8 +100,7 @@ void tcp_connection::read_loop_async() {
   });
 }
 
-// !!COPY
-void tcp_connection::write_async(std::vector<char> message_) {
+void tcp_connection::write_async(const std::vector<char>& message_) {
```
NikoAri, line 103: >r>& m [](start = 54, length = 5)

Why not just have Api take &&?

It's being moved anyways on line 107
Also, why isn't there an error/warning when move is performed on const&?

src/etw/tracing.cpp
```diff
@@ -0,0 +1,24 @@
+#include "tracing.h"
+
+// Define the GUID to use in TraceLoggingProviderRegister
+// {85A99459-1F1D-49BD-B3DC-E5EF2AD0C2C8}
+TRACELOGGING_DEFINE_PROVIDER(g_hV8JSIRuntimeTraceLoggingProvider,
+                             "Microsoft.V8JSIRuntime",
+                             (0x85a99459, 0x1f1d, 0x49bd, 0xb3, 0xdc, 0xe5,
```
NikoAri, line 7: >(0x [](start = 29, length = 3)

(nit) isn't there some simple Guid conversion to get this from string?

scripts/tracing/trace.ps1
```diff
@@ -0,0 +1,118 @@
+<#
+  .SYNOPSIS
+  Starts a Trace recording session enabling Hermes TraceLogging provider.
```
tudorms, line 3: >Hermes [](start = 44, length = 6)

(Nit) V8

src/V8JsiRuntime.cpp
```diff
@@ -704,7 +762,13 @@ jsi::Value V8Runtime::ExecuteString(
       v8::String::NewFromUtf8(
           isolate, reinterpret_cast<const char *>(sourceURL.c_str()))
           .ToLocalChecked();
+  /*v8::ScriptOrigin origin(
```
NikoAri, line 765: > /*v8: [](start = 1, length = 6)

Remove

src/V8Platform.cpp
```diff
@@ -158,7 +151,7 @@ void ETWTracingController::UpdateTraceEventDuration(
     const char *name,
     uint64_t handle) {
 #if defined(_WIN32) && !defined(__clang__)
-  EventWriteUPDATE_TIMESTAMP(name, handle);
+  //EventWriteUPDATE_TIMESTAMP(name, handle);
```
NikoAri, line 154: >//Ev [](start = 2, length = 4)

Remove

NikoAri (approved):
:shipit:

src/V8JsiRuntime.cpp
```diff
@@ -704,7 +762,13 @@ jsi::Value V8Runtime::ExecuteString(
       v8::String::NewFromUtf8(
           isolate, reinterpret_cast<const char *>(sourceURL.c_str()))
           .ToLocalChecked();
+  /*v8::ScriptOrigin origin(
+      urlV8String, 0, 0, false, -1,
+      v8::String::NewFromUtf8(isolate,
+          reinterpret_cast<const char*>("file:///E:/github/fibonacci/dist/main.js.map"))
+          .ToLocalChecked());*/
```
tudorms, line 769: remove?

src/V8JsiRuntime.cpp
```diff
@@ -1307,14 +1375,60 @@ jsi::HostFunctionType &V8Runtime::getHostFunction(const jsi::Function &obj) {
...
     const jsi::Value &jsThis,
     const jsi::Value *args,
     size_t count) {
+ 
   _ISOLATE_CONTEXT_ENTER
   v8::Local<v8::Function> func =
       v8::Local<v8::Function>::Cast(objectRef(jsiFunc));
+
+  // TODO: This may be a bit costly when there are large number of JS function
+  // calls from native. Evaluate !
```
tudorms, line 1417: >Evaluate [](start = 24, length = 8)

Could we comment out this and the traces until we can evaluate the cost?

src/public/V8JsiRuntime.h
```diff
@@ -78,17 +78,36 @@ struct V8RuntimeArgs {
...
+  // Enabling all the diagnostic and tracings by default so that we will be able to tune them and writing tools over them.
+  bool trackGCObjectStats{true};
+  bool enableTracing{true};
+  bool enableJitTracing{true};
+  bool enableMessageTracing{true};
+  bool enableLog{true};
+  bool enableGCTracing{true};
+
+  // Enabling inspector by default. This will help in stabilizing inspector, and easily debug JS code when needed.
+  // There shouldn't be any perf impacts until a debugger client is attached, except the overload of having a WebSocket port open, which should be very small.
```
tudorms, line 90: >which should be very small [](start = 131, length = 26)

I think we shouldn't enable this by default for ship builds. It feels weird to open a port on customer's machines and could cause issues and questions.

src/inspector/inspector_socket.cpp
mganandraj: Not sure .. Do we get Office logs with the Watson dump ? Or do we have to go to Cosmos and query for session logs ?

Vlad mentioned long time that there is a process in Windows to enable specific traces in user machines and collect it when needed .. But it was not retrospective collection for sure.

src/V8JsiRuntime.cpp
mganandraj: Tudor, I think we should keep it and measure it through our tooling. This is not supposed to be expensive as both calls to get the function name only does lookups. My main concern is many conditional statements. 
Having these traces will really help in analyzing issues in future as these give us the function call context.

src/etw/tracing.cpp
mganandraj: I tried something similar with some macros such as __uuid, UUID etc. but i didn't have success. If you know one please share so that i can update the code.

src/inspector/inspector_tcp.cpp
mganandraj: Yeah .. It should be a move all the way ..

PR #47, opened 2021-03-25
Implement ArrayBuffer runtime methods

rgerd:
[BabylonReactNative](https://github.com/BabylonJS/BabylonReactNative) uses a [BabylonNative polyfill](https://github.com/BabylonJS/BabylonNative/blob/master/Polyfills/XMLHttpRequest/Source/XMLHttpRequest.cpp#L108) to request a remote resource, and that [uses JSI ArrayBuffer methods](https://github.com/BabylonJS/BabylonNative/blob/master/Dependencies/napi/napi-jsi/include/napi/napi-inl.h#L789), which then call into our implementation of the V8 JSI runtime.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/47)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.10",
+    "version": "0.64.11",
     "v8ref": "refs/branch-heads/9.0"
 }
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -1239,16 +1239,19 @@ bool V8Runtime::isArray(const jsi::Object &obj) const {
   return objectRef(obj)->IsArray();
 }
 
-bool V8Runtime::isArrayBuffer(const jsi::Object & /*obj*/) const {
-  throw std::runtime_error("Unsupported");
+bool V8Runtime::isArrayBuffer(const jsi::Object &obj) const {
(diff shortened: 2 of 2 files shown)


tudorms (approved):
This looks great, thank you!

Please also bump the version in \config.json (otherwise a new NuGet won't be generated).

JSI doesn't have a way to create an ArrayBuffer from the native side, so we thought this functionality is just not complete yet; I presume you create the ArrayBuffers somehow JS-side and only need to read them in native?

rgerd:
> JSI doesn't have a way to create an ArrayBuffer from the native side, so we thought this functionality is just not complete yet; I presume you create the ArrayBuffers somehow JS-side and only need to read them in native?

@tudorms That's correct. We [get the ArrayBuffer javascript constructor here](https://github.com/BabylonJS/BabylonNative/blob/e56f0ce041087be7d693524ff024252b134f760d/Dependencies/napi/napi-jsi/include/napi/napi.h#L48) and then [use it here to create the ArrayBuffer on the JS side](https://github.com/BabylonJS/BabylonNative/blob/master/Dependencies/napi/napi-jsi/include/napi/napi-inl.h#L788).

PR #49, opened 2021-04-01
Inspector improvements

mganandraj:
This PR makes the following enhancements to the Inspector,
1. Support multiple isolates on single server/port
2. Cleanly close the connection when the V8JSI runtime tears down. This should fix the crash on reload. (There is still a crash if you kill/close the app when an active inspector session exists. I'm still looking into it.)
3. CPU profile is more reliable
4. Inspector traces created from JS thread which make them a little more easier to consume.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/49)


tudorms (approved):
:shipit:

PR #50, opened 2021-04-01
Tracing improvements

mganandraj:
This PR makes the following improvement to the ETW tracing,
1. All traces are now correctly attributed to the right runtime.
2. Fixes crash when multiple runtimes are registering the ETW provider.
3. Enables overriding some arguments through system registry.
4. Improvements to the powershell script to make the tracing easier.

I reformatted some of the files accidentally. I can't find an easy way to revert those formatting hcanges.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/50)


src/public/V8JsiRuntime.h
```diff
@@ -76,13 +76,11 @@ struct V8RuntimeArgs {
 
   bool backgroundMode{false};
 
-  // Enabling all the diagnostic and tracings by default so that we will be able to tune them and writing tools over them.
-  bool trackGCObjectStats{true};
-  bool enableTracing{true};
-  bool enableJitTracing{true};
-  bool enableMessageTracing{true};
-  bool enableLog{true};
```
tudorms, line 84: >bool enableLog{true}; [](start = 2, length = 21)

Can you keep enableLog in there even if not used (maybe with a comment)? Otherwise, the change will break the ABI meaning we'd need to recompile all dependencies up-the-chain from source instead of just dropping a new DLL directly in the consumer.

tudorms:
Please also bump the version in config.json

src/V8JsiRuntime.cpp
```diff
@@ -1,24 +1,22 @@
...
-#include "V8Platform.h"
-#include "public/ScriptStore.h"
-
 #include <atomic>
 #include <cstdlib>
 #include <list>
 #include <mutex>
 #include <sstream>
 
+#include "V8JsiRuntime_impl.h"
```
vmoroz, line 9: We typically want the main .h file to be the first one. It is also better to put all standard headers after custom header - this way we can partially ensure that custom headers have all required includes.

vmoroz:
It is difficult to review changes because of the formatting changes in the same PR.
Please split the PR into two different PRs: 
- The first one one will be the formatting changes only. Note that the code is not foormatted using the clang format settings that we have. Open the folder in VS Code and format there - it should pickup the Clang format settings.
- Then after the formatting change is merged - reformat your code using Clang format, and merge the previous PR. - After that we should see only your changes.

src/V8Platform.h
```diff
@@ -28,52 +28,37 @@ class ETWTracingController : public v8::TracingController {
   ETWTracingController(bool enabled) : enabled_(enabled) {}
   ~ETWTracingController() = default;
 
-  const uint8_t *GetCategoryGroupEnabled(const char *category_group) override;
+  const uint8_t* GetCategoryGroupEnabled(const char* category_group) override;
```
vmoroz, line 31: >const char*  [](start = 41, length = 12)

Our clang format settings put the star on the right. What tool do you use to format code?

src/V8JsiRuntime.cpp
```diff
@@ -1574,12 +1572,54 @@ v8::Local<v8::Value> V8Runtime::valueRef(const jsi::Value &value) {
   }
 }
 
-std::unique_ptr<jsi::Runtime> makeV8Runtime(V8RuntimeArgs &&args) {
+void applyOverrides(V8RuntimeArgs& args) {
```
vmoroz, line 1575: >applyOverrides [](start = 5, length = 14)

It would be great to avoid this DLL accessing the registry. Can it be the application that does it?
The V8JSI is an open source project and it seems to be wrong to use MS Office key in the path and have a "backdoor" affecting any app in the system that uses this DLL.

tudorms:
Alternatively just reformat the code in vscode / clang format and push a PR interation here.

---
In reply to: [812684252](https://github.com/microsoft/v8-jsi/pull/50#issuecomment-812684252) [](ancestors = 812684252)

mganandraj:
> It is difficult to review changes because of the formatting changes in the same PR.
> Please split the PR into two different PRs:
> 
> * The first one one will be the formatting changes only. Note that the code is not foormatted using the clang format settings that we have. Open the folder in VS Code and format there - it should pickup the Clang format settings.
> * Then after the formatting change is merged - reformat your code using Clang format, and merge the previous PR. - After that we should see only your changes.

The code got (accidentally) formatted by VS using chromium formatting rules .. I realized the formatting changes after making a bunch of changes on top, and attempted reverting but it turned out to be very messy .. 

I don't have strong opinions on the formatting rules. When we started this codebase, we roughly followed the V8 formatting rules for things such as the variable, function naming, which still stays. As we build the code alongside V8, and devs (incl. me) usually create one big VS solution of V8 with this JSI code as a subproject, the chromium formatting rules fits the best i think. But, that is a very subjective discussion which I don't want to get into.

I'm reformatting the code with VS code using our checked in clang format as @tudorms suggested which is the least expensive resolution. Hope it works for you.

@tudorms

src/V8JsiRuntime.cpp
mganandraj: I do partially agree Vladimir .. but having such control will be extremely helpful when building debugging and performance tooling. We can hopefully remove these once our tooling is matured. I'm removing the "Office" subpaths from the registry key.. Hope it works for you.

src/V8JsiRuntime.cpp
```diff
@@ -28,7 +26,9 @@ using namespace facebook;
 
 namespace v8runtime {
 
-thread_local uint16_t V8Runtime::tls_isolate_usage_counter_ = 0;
+thread_local uint16_t V8Runtime::tls_isolate_usage_counter_{0};
+thread_local std::weak_ptr<V8Runtime::CounterStores>
```
vmoroz, line 30: >weak_ptr [](start = 18, length = 8)

Ideally we should stop using TLS. In future we want to be able to use The V8 engine in scenarios without a dedicated JS thread. A better approach is to assciate the data with the Runtime, V8 context, or V8 isolate. 
Anyway, there is no mechanism to call destructor on the TLS stored objects. Please reconsider having std::weak_ptr as TLS value. TLS is usually good only for simple values such as numbers or raw pointers.

vmoroz: I looked though the code and still do not understand why we cannot associate counters with the Runtime or Isolate instead of TLS.

---
In reply to: [606609468](https://github.com/microsoft/v8-jsi/pull/50#discussion_r606609468) [](ancestors = 606609468)

src/V8JsiRuntime.cpp
vmoroz: Can we have a subkey per application? I.e. per executable file name that uses the V8JSI DLL.

---
In reply to: [606578849](https://github.com/microsoft/v8-jsi/pull/50#discussion_r606578849) [](ancestors = 606578849)

src/V8JsiRuntime.cpp
```diff
@@ -572,19 +589,22 @@ void V8Runtime::initializeV8() {
 }
 
 V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
+  counter_stores_ = std::make_shared<CounterStores>();
+  tls_counter_stores_ = counter_stores_;
```
vmoroz, line 593: Why do we create them in two places: here in constructor and in the CreateNewIsolate() method? In most cases the CreateNewIsolate will override values created here.

mganandraj: Ah .. This was a mistake .. Thanks !

src/V8JsiRuntime.cpp
mganandraj: We need to access the counters from the static callback functions that we register to the isolate .. Essentially, what we want is access to "this" runtime pointer in the static callbacks .. The access to counters starts even before the isolate is properly constructed, hence we can't use the isolate data slots .. 

I was also hesistant to add tls data .. I initially had a unique instance id counter to uniquely identify an instance, instead of a thread id .. But it turned out to be a lot more code and uglier. .. As V8 isolates and a thread is 1:1 mapped, I think the tls is fine for our scenario.. Essentialy, we can treat the counters as per isolate counters, not really per-runtime counters ..

We are not really gaining anything by having the weak_ptr .. even a raw pointer in tls with the shared_ptr member for lifetime manager suffice for us .. But, i made this change to weak_ptr just before the CR to make the code look nicer :) the naked pointer looks bad .. But as you suggested this, i'll revert the tls entry to a nake pointer.

src/V8JsiRuntime.cpp
tudorms: Why is this needed at all? These values are already being passed through the arguments; the consumer is already capable of passing them in.

---
In reply to: [606610175](https://github.com/microsoft/v8-jsi/pull/50#discussion_r606610175) [](ancestors = 606610175,606578849)

src/V8JsiRuntime.cpp
```diff
@@ -1574,7 +1667,69 @@ v8::Local<v8::Value> V8Runtime::valueRef(const jsi::Value &value) {
...
+    args.trackGCObjectStats = enable;
+    args.enableTracing = enable;
+    args.enableJitTracing = enable;
+    args.enableMessageTracing = enable;
+    args.enableGCTracing = enable;
+  }
+}
+
 std::unique_ptr<jsi::Runtime> makeV8Runtime(V8RuntimeArgs &&args) {
+  applyOverrides(args);
```
tudorms, line 1732: >applyOverrides(args); [](http://example.com/codeflow?start=2&length=21)

Can you submit the rest of the changes besides this registry functionality (I think this is the only sticking point of the PR and we can revisit it later). There's a formatting PR in the pipeline and merging that would be painful after the fact.

vmoroz:
This PR is outdated. We may return to these ideas later, but for now we are closing it.

PR #51, opened 2021-04-07
Implement N-API for V8

vmoroz:
This PR contains the initial implementation of N-API for V8 in the V8JSI.DLL.

The N-API is the ABI-safe "C"-based JavaScript engine API implemented by Node.js for extension modules.
It is documented here: https://nodejs.org/api/n-api.html
We believe that we must use it as a basis for the ABI-safe JavaScript engine API instead of inventing our own from scratch.

The N-API for V8 implementation and the unit tests are taken from the Node.js project code: https://github.com/nodejs/node.
We try to keep the original files unchanged. The Node.js original files were added in this PR's branch as the first commit: https://github.com/microsoft/v8-jsi/commit/7da7426c7d915cdb781010ea2a85f5dac3ec2e7c

On top of the Node.js code we did the following changes:
- Added new public APIs napi_ext_* to host the JavaScript engine.
- Make sure that we have only minimal subset of Node.js code that the N-API for V8 implementation depends on.
- Changed V8JsiRuntime to be used as an internal N-API environment object. (The public `napi_env__` is implemented by Node.js code.)
- Run unit JavaScript-based Node.js N-API unit test as GTests. We simulate the Node.js execution of the tests.
- Added Readme.md files with notes about the borrowed code and the respective copyright notes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/51)


NikoAri:
> #include <climits>  // INT_MAX

(nit) we should probably add copyright note to all of the files

---
Refers to: src/napi/js_native_api_v8.cc:1 in a4e3d45. [](commit_id = a4e3d457ddfe20c179c3dc0010a5d42058d1c9ee, deletion_comment = False)

NikoAri (approved):
:shipit:

vmoroz:
> > #include   // INT_MAX
> 
> (nit) we should probably add copyright note to all of the files
> 
> Refers to: src/napi/js_native_api_v8.cc:1 in [a4e3d45](https://github.com/microsoft/v8-jsi/commit/a4e3d457ddfe20c179c3dc0010a5d42058d1c9ee). [](commit_id = [a4e3d45](https://github.com/microsoft/v8-jsi/commit/a4e3d457ddfe20c179c3dc0010a5d42058d1c9ee), deletion_comment = False)

This file is from the Node.js code. I wan to keep it unchanged. Instead, I have put all the related copyright attributions to the Readme.md

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+    args.ignoreUnhandledPromises = true;
+  }
+
+  auto runtime = std::make_unique<v8runtime::V8Runtime>(std::move(args));
+
+  auto context = v8impl::PersistentToLocal::Strong(runtime->GetContext());
+  *env = new napi_env__(context);
+
+  // Let the runtime exists. It can be accessed from the Context.
+  runtime.release();
```
mganandraj, line 105: >runtime [](start = 2, length = 7)

Do we plan to share a runtime between React native and node modules or any other consumers of napi ?

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+#include "env-inl.h"
+
+#include "V8JsiRuntime_impl.h"
+#include "js_native_api_v8.h"
+#include "js_native_ext_api.h"
+#include "public/ScriptStore.h"
```
mganandraj, line 35: >"public/ScriptStore.h" [](start = 9, length = 22)

I think this include is unnecessary.

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+      *info.function ? ":" : "",
+      info.message);
+  fflush(stderr);
+
+  // [vmoroz] Abort();
+}
+
+} // namespace node
+
+// TODO: [vmoroz] verify that finalize_cb runs in JS thread
```
mganandraj, line 229: >finalize_cb runs in JS [](start = 30, length = 22)

As this is triggered just before finalizing the object, it is highly likely to be called synchrously from the thread where the GC is taking place, which could possibly be one of the background threads.

src/V8JsiRuntime.cpp
```diff
@@ -593,15 +624,21 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
...
+  auto context = CreateContext(isolate_);
+  context_.Reset(isolate_, context);
+
+  // Associate the Runtime with the V8 context
+  context->SetAlignedPointerInEmbedderData(ContextEmbedderIndex::Runtime, this);
+  // Used by Environment::GetCurrent to know that we are on a node context.
+  context->SetAlignedPointerInEmbedderData(
+      ContextEmbedderIndex::ContextTag, V8Runtime::RuntimeContextTagPtr);
+
+  v8::Context::Scope context_scope(context);
```
mganandraj, line 638: We ideally want to allow multiple contexts in the same thread sharing the same isolate .. This will break it.

src/V8JsiRuntime.cpp
```diff
@@ -1574,6 +1673,106 @@ v8::Local<v8::Value> V8Runtime::valueRef(const jsi::Value &value) {
...
+  return static_cast<V8Runtime *>(context->GetAlignedPointerFromEmbedderData(
+      ContextEmbedderIndex::Runtime));
+}
+
+bool V8Runtime::HasUnhandledPromiseRejection() noexcept {
+  return !!last_unhandled_promise_;
+}
+
+std::unique_ptr<UnhandledPromiseRejection>
+V8Runtime::GetAndClearLastUnhandledPromiseRejection() noexcept {
```
mganandraj, line 1702: >GetAndClearLastUnhandledPromiseRejection [](start = 11, length = 40)

Vladimir, i'm curious to know the motivation behind adding these .. Are there to get some tests or any specific node modules working ?

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+namespace per_process {
+// From node.cc
+// Tells whether the per-process V8::Initialize() is called and
+// if it is safe to call v8::Isolate::GetCurrent().
+bool v8_initialized = false;
+} // namespace per_process
+
+// From node_errors.cc
+[[noreturn]] void Assert(const AssertionInfo &info) {
+  std::string name = "aaa"; // [vmoroz] GetHumanReadableProcessName();
```
mganandraj, line 212: >"aaa"; // [vmoroz] GetHumanReadableProcessName(); [](start = 21, length = 49)

In inspector, we used something like "VsJsiHost"

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+  fprintf(
+      stderr,
+      "%s: %s:%s%s Assertion `%s' failed.\n",
+      name.c_str(),
+      info.file_line,
+      info.function,
+      *info.function ? ":" : "",
+      info.message);
+  fflush(stderr);
```
mganandraj, line 222: TRACEV8RUNTIME_CRITICAL ?

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+  // [vmoroz] Abort();
```
mganandraj, line 224: >// [vmoroz] Abort(); [](start = 2, length = 20)

std::abort() ?

src/napi/js_native_ext_api_v8.cpp
```diff
@@ -0,0 +1,273 @@
...
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+#include "env-inl.h"
+
+#include "V8JsiRuntime_impl.h"
+#include "js_native_api_v8.h"
+#include "js_native_ext_api.h"
+#include "public/ScriptStore.h"
+
+static struct napi_env_scope__ {
```
mganandraj, line 37: >napi_env_scope__ [](start = 14, length = 16)

I like our current approach of creating these scope objects in stack better than this..

vmoroz: Could you please explain what do you mean by that?
The C-based API cannot have stack-based objects because C does not have classes and destructors.
These C-based must not be used directly in C++ for exception-safety, they must be wrapped up into C++ classes that we put on the call stack which will automatically open/close the scope.

src/V8JsiRuntime.cpp
vmoroz: Without this code we cannot observe any errors which happen in the Promise continuations. As an alternative we can consider custom implementation of the V8 microtasks, I just did not try it yet. What is your concern here?

src/V8JsiRuntime.cpp
vmoroz: Could you explain how it breaks it? This code associates a context with the Runtime. So, we can always retrieve the runtime from the context. This code does not do anything related to a thread.
The thread affinity is introduced by the TLS variables that we have and are continuing to add. The TLS variables must be removed in future and replaced with this mechanism where we can easily get from the Runtime to context and vice versa.
The next step will be to have APIs such as napi_ext_open_env_scope/napi_ext_close_env_scope that bind the Runtime to a specific thread while the Runtime is in scope.

src/V8JsiRuntime.cpp
mganandraj: As i mentioned, my comment was more our of curiosity than a concern, that why is this needed now for napi .. I completely welcome having this code, but was not sure how will this be used .. we didn't need it with react native as wedon't use engine native promises in RN environment.. 

---
In reply to: [611882322](https://github.com/microsoft/v8-jsi/pull/51#discussion_r611882322) [](ancestors = 611882322)

src/napi/js_native_ext_api_v8.cpp
vmoroz: I need to write a unit test here and if the thread assumption is broken, then I need to have a different implementation.

src/napi/js_native_ext_api_v8.cpp
vmoroz: The V8JsiRuntime.h uses ScriptStore.h types without including the ScriptStore.h. If I do not include ScriptStore.h here, then I get the linker error while trying to use the V8Runtime creation arguments.

src/V8JsiRuntime.cpp
mganandraj: Ah .. I see.. When reading this last time, i thought this scope object is kept with the runtime .. Here, you are only creating a stack variable.. Im curious to konw what is the purpose of this statement ? No new code seems to be added in the constructor after this statement.

---
In reply to: [611886893](https://github.com/microsoft/v8-jsi/pull/51#discussion_r611886893) [](ancestors = 611886893)

src/V8JsiRuntime.cpp
vmoroz: I needed it for the unit tests to run. Without this code I could not observe errors or asserts in Promise continuations. It is not part of the official NAPI API. I had to add an extension to enable it. Though, I am not 100% satisfied with this new APIs. I want to have something more fundamental and to be able to use it for other JS engines.

src/V8JsiRuntime.cpp
```diff
@@ -69,27 +76,31 @@ class TaskRunnerAdapter : public v8::TaskRunner {
       : taskRunner_(std::move(taskRunner)) {}
 
   void PostTask(std::unique_ptr<v8::Task> task) override {
-    taskRunner_->postTask(std::make_unique<TaskAdapter>(std::move(task)));
+    taskRunner_->postTask(node::static_unique_pointer_cast<JSITask>(
```
tudorms, line 79: >node::static_unique_pointer_cast<JSITask> [](start = 26, length = 41)

Is this needed? I think you noticed this code is not actually used.

vmoroz: The VS Code IntelliSense was unhappy about it and I saw the opportunity to use the Node.js function to fix it.

src/napi/js_native_ext_api_v8.cpp
vmoroz: I will resolve this TODO in a future PR

src/V8JsiRuntime.cpp
vmoroz: Please note that I did not add this statement: it is the previously existing code `v8::Context::Scope context_scope(context_.Get(GetIsolate()));` - it was simplified to use the `context` variable. If I remove it, then tests continue to work, but I want to keep it for now because I do not know all existing scenarios.

---
In reply to: [611890075](https://github.com/microsoft/v8-jsi/pull/51#discussion_r611890075) [](ancestors = 611890075,611886893)

src/napi/js_native_ext_api_v8.cpp
vmoroz: Replaced with _get_pgmptr call

src/napi/js_native_ext_api_v8.cpp
vmoroz: I have added it, but we may add more parameters there.

src/napi/js_native_ext_api_v8.cpp
vmoroz: I put std::terminate there.

tudorms (approved):
:shipit:

PR #52, opened 2021-04-15
Set Clang format ColumnLimit to 120

vmoroz:
Last few PRs have a lot of code changes mostly because we had to auto-format the code.
In this PR we format existing code files and set the Clang format ColumnLimit to 120 as in the react-native-windows projects.
It would be ideal if this PR is merged before the other two outstanding PRs to reduce number of changed lines in these PRs.
There are no code changes in PR besides the auto-formatting and fixing some spelling issues in a few comments.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/52)

### DIFF
--- a/src/V8Platform.h
+++ b/src/V8Platform.h
@@ -59,16 +59,11 @@ class ETWTracingController : public v8::TracingController {
       unsigned int flags,
       int64_t timestamp) override;
 
-  void UpdateTraceEventDuration(
-      const uint8_t *category_enabled_flag,
-      const char *name,
-      uint64_t handle) override;
+  void UpdateTraceEventDuration(const uint8_t *category_enabled_flag, const char *name, uint64_t handle) override;
 
-  void AddTraceStateObserver(
-      v8::TracingController::TraceStateObserver *observer) override;
+  void AddTraceStateObserver(v8::TracingController::TraceStateObserver *observer) override;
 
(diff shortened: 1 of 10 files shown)


src/testmain.cpp
```diff
@@ -16,7 +16,8 @@ std::vector<facebook::jsi::RuntimeFactory> runtimeGenerators() {
   }};
 }
 
-}}
```
NikoAri, line 19: (nit) could we use modern style namespace::syntax?

src/V8Platform.h
```diff
@@ -59,16 +59,11 @@ class ETWTracingController : public v8::TracingController {
       unsigned int flags,
       int64_t timestamp) override;
 
-  void UpdateTraceEventDuration(
-      const uint8_t *category_enabled_flag,
```
NikoAri, line 63: > co [](start = 5, length = 3)

(nit) usually, having separate line per argument for long arguments is more readable

NikoAri (approved):
:shipit:

src/V8Platform.h
vmoroz: This is an automatic formatting. We better keep it "as-is".

PR #53, opened 2021-04-21
Implement JSI on top of N-API

vmoroz:
The new JSI implementation based on N-API is the new public files NapiJsiRuntime.h and NapiJsiRuntime.impl.h.
The implementation is passing all JSI unit tests (added in the testmain.cpp).

Other related changes:
- Some APIs are changed in the jsi_native_ext_api.h
- The new compat.h public file has definitions for string_view and span types defined in C++17 and C++20.
- The Windows.h include is extracted into the new V8Windows.h. All files that used Windows.h are changed to use the new V8Windows.h.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/53)


tudorms (approved):
:shipit:

PR #54, opened 2021-04-27
Use ASIO instead of Boost

tudorms:
This removes the dependency on Boost, requiring the (much smaller) ASIO library instead.

Also includes the update to V8 version 9.1.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/54)


vmoroz (approved):
:shipit:

PR #57, opened 2021-05-28
Disable web assembly; add perf tuning flags

tudorms:
- Disable webassembly support (this shaves 2Mb of the DLL size in x64 release)
- Expose V8 performance tuning flags through the arguments
- Make the flags in V8RuntimeArgs more flexible and resilient to ABI changes (by using a padded struct with some leftover room)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/57)

### DIFF
--- a/src/public/V8JsiRuntime.h
+++ b/src/public/V8JsiRuntime.h
@@ -26,17 +26,10 @@ struct JSITaskRunner {
 };
 
 struct V8RuntimeArgs {
+  // This is only used by the debugger: inspector needs to wake up the js thread for message dispatching
+  // It cannot be done in an asynchronous lambda because V8Inspector internally uses "main thread" handles
   std::shared_ptr<JSITaskRunner> foreground_task_runner; // foreground === js_thread => sequential
   std::unique_ptr<facebook::jsi::PreparedScriptStore> preparedScriptStore;
-
-  // Enabling all the diagnostic and tracings by default so that we will be able to tune them and writing tools over
-  // them.
-  bool trackGCObjectStats{true};
-  bool enableJitTracing{true};
-  bool enableMessageTracing{true};
-  bool enableGCTracing{true};
-  bool enableInspector{false};
-  bool waitForDebugger{false};
 
   // To debug using vscode-node-adapter create a blank vscode workspace with the following launch.config and attach to
   // the runtime.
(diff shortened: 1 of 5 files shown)


src/public/V8JsiRuntime.h
```diff
@@ -61,8 +54,27 @@ struct V8RuntimeArgs {
   size_t initial_heap_size_in_bytes{0};
   size_t maximum_heap_size_in_bytes{0};
 
-  bool enableGCApi{false};
-  bool ignoreUnhandledPromises{false};
+  // Padded to allow adding boolean flags without breaking the ABI
```
vmoroz, line 57: >adding boolean flags [](http://example.com/codeflow?start=21&length=20)

I like the idea about the bit flags!

src/public/V8JsiRuntime.h
```diff
@@ -61,8 +54,27 @@ struct V8RuntimeArgs {
   size_t initial_heap_size_in_bytes{0};
   size_t maximum_heap_size_in_bytes{0};
 
-  bool enableGCApi{false};
-  bool ignoreUnhandledPromises{false};
+  // Padded to allow adding boolean flags without breaking the ABI
+  union {
+    struct {
```
vmoroz, line 59: >struct { [](http://example.com/codeflow?start=4&length=8)

I do not think that the anonymous structs are the standard C++.

vmoroz (approved):
:shipit:

src/public/V8JsiRuntime.h
tudorms: I tried a few compilers in godbolt and they all like the anonymous structs. The only thing not working until c++20 is default initialization for bitfields in anonymous structs (so if we need any of these flags to default to something other than false, we'd need a constructor).

vmoroz: Yes, in practice all most common compilers like MSVC, Clang, and GCC support it, but it is not part of the standard. E.g. see https://docs.microsoft.com/en-us/cpp/cpp/anonymous-class-types?view=msvc-160, https://stackoverflow.com/questions/16202576/c-anonymous-structs

vmoroz: The MSVC fails to compile the anonymous structs if it is compiled with "Disable Language Extensions: Yes (/Za)".
I would suggest to name the struct field. E.g. `flags`.

tudorms: Makes sense, I'll do that. Thanks!

PR #59, opened 2021-06-14
Enable open inspector servers on active runtime instance

mganandraj:
This change has the following effects,
1. API available to open inspector server on an active runtime instance.
2. Ensure that the server management APIs in inspector can be called from any thread.
3. Ensuring the contexts are registered correctly to the inspector. This will correct some issues in multi context runtimes, and when sharing isolates between runtimes, and also result in more symmetric APIs. I don't think we have a multi context/shared-isolate scenario in Office, but having the right model will help us in better reasoning when diagnosing issues.
4. Some improvements in the shutdown flow.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/59)

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -515,10 +515,20 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
   v8::Context::Scope context_scope(context);
 
 #if defined(_WIN32) && defined(V8JSI_ENABLE_INSPECTOR)
+  void* inspector_agent = isolate_->GetData(ISOLATE_INSPECTOR_SLOT);
+  if (inspector_agent) {
+    inspector_agent_ = reinterpret_cast<inspector::Agent*>(inspector_agent)->getShared();
+  } else {
+    inspector_agent_ = std::make_shared<inspector::Agent>(
+        isolate_, args_.inspectorPort);
+    isolate_->SetData(ISOLATE_INSPECTOR_SLOT, inspector_agent_.get());
+  }
+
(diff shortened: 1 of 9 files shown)


src/V8JsiRuntime.cpp
```diff
@@ -515,10 +515,20 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
   v8::Context::Scope context_scope(context);
 
 #if defined(_WIN32) && defined(V8JSI_ENABLE_INSPECTOR)
+  void* inspector_agent = isolate_->GetData(ISOLATE_INSPECTOR_SLOT);
+  if (inspector_agent) {
+    inspector_agent_ = reinterpret_cast<inspector::Agent*>(inspector_agent)->getShared();
+  } else {
+    inspector_agent_ = std::make_shared<inspector::Agent>(
```
tudorms, line 522: >inspector_agent_ [](http://example.com/codeflow?start=4&length=16)

I presume this doesn't create network sockets until started, right? The memory impact of always creating an inspector agent for any runtime should be pretty minimal but it might be worth taking a quick look

tudorms (approved):
:shipit:

src/V8JsiRuntime.cpp
mganandraj: Thanks Tudor for taking a quick look.. 
Yes, we are now only initializing the inspector client, not starting the network sockets. I've done some preliminary analysis on the performance impacts.

PR #60, opened 2021-06-15
Revert webassembly disabling

tudorms:
- revert webassembly disabling which was causing issues with x86 builds (this needs further investigation and follow-up with Google);
- expose debuggerRuntimeName from arguments;

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/60)

### DIFF
--- a/src/public/V8JsiRuntime.h
+++ b/src/public/V8JsiRuntime.h
@@ -54,6 +54,9 @@ struct V8RuntimeArgs {
   size_t initial_heap_size_in_bytes{0};
   size_t maximum_heap_size_in_bytes{0};
 
+  // Set this to override the target name displayed in the debugger (to distinguish multiple parallel runtimes)
+  std::string debuggerRuntimeName;
+
   // Padded to allow adding boolean flags without breaking the ABI
   union {
     struct {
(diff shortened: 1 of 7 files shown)


src/public/V8JsiRuntime.h
```diff
@@ -54,6 +54,9 @@ struct V8RuntimeArgs {
   size_t initial_heap_size_in_bytes{0};
   size_t maximum_heap_size_in_bytes{0};
 
+  // Set this to override the target name displayed in the debugger (to distinguish multiple parallel runtimes)
+  std::string debuggerRuntimeName;
```
NikoAri, line 58: >debuggerRuntimeName [](http://example.com/codeflow?start=14&length=19)

(nit) where is this being set?

NikoAri (approved):
:shipit:

src/public/V8JsiRuntime.h
NikoAri: Or there will be a separate CR to set it in RN?

tudorms: Yes, this needs to come in from external code using the V8Runtime (like RNW or devmain)

PR #62, opened 2021-06-25
Remove the dbghelp dependency

tudorms:
dbghelp.dll is only used for symbol resolution if using V8's unhandled exception filter. We don't make much use of that and it makes it harder to port to ARM64EC, so removing the dependency.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/62)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.63.9",
+    "version": "0.63.10",
     "v8ref": "refs/branch-heads/9.1"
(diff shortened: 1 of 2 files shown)


NikoAri (approved):
:shipit:

Khalef1 (approved):

PR #70, opened 2021-08-10
Ensuring safe Inspector tear down

mganandraj:
This change is ensuring that the inspector session gets torn down when the runtime is torn down. We accidentally removed some code in one of the prioer change which is being reinstated, and in addition we are also ensuring that the server itself is torn down when all the connected target are torn down.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/70)


src/inspector/inspector_agent.cpp
```diff
@@ -393,40 +388,57 @@ void InspectorWrapConsoleCall(const v8::FunctionCallbackInfo<v8::Value> &args) {
...
-  
-  return *server_;
-}
+
+  void removeServer(int port) {
+    servers_.erase(port);
+  }
+
+  static ServerCollection& Instance() {
+    static ServerCollection instance;
```
NikoAri, line 427: >instance [](http://example.com/codeflow?start=28&length=8)

(nit) can we add prefix "s_"?

src/inspector/inspector_agent.cpp
```diff
@@ -393,40 +388,57 @@ void InspectorWrapConsoleCall(const v8::FunctionCallbackInfo<v8::Value> &args) {
...
+    return instance;
+  }
+
+ private:
+  std::mutex mutex_server_init_;
+  std::unordered_map<int, std::weak_ptr<inspector::InspectorSocketServer>> servers_;
```
NikoAri, line 433: >int [](http://example.com/codeflow?start=21&length=3)

(nit) is this port? could you add comment like:
std::unordered_map<int / *port* /, std::we

src/inspector/inspector_tcp.cpp
```diff
@@ -21,7 +21,14 @@ tcp_server::tcp_server(int port, ConnectionCallback callback, void* data)
 }
 
 void tcp_server::run() {
-  io_service_.run();
+  try {
+    io_service_.run();
+  } catch (std::exception const& e) {
```
NikoAri, line 26: >catch [](http://example.com/codeflow?start=4&length=5)

(nit) should we catch general exception as well? eg:
catch (...)

NikoAri (approved):
:shipit:

config.json
```diff
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.20",
-    "v8ref": "refs/branch-heads/9.2"
+    "version": "0.64.21",
+    "v8ref": "refs/branch-heads/9.1"
```
tudorms, line 3: >9.1 [](http://example.com/codeflow?start=32&length=3)

This was probably a bad merge, we want 9.2 (or 9.3 if the new PR lands first)

tudorms (approved):
:shipit:

PR #71, opened 2021-08-10
V8 9.3

tudorms:
Integrated the [newly released](https://v8.dev/blog/v8-release-93) 9.3 update.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/71)

### DIFF
--- a/.ado/pull-request.yml
+++ b/.ado/pull-request.yml
@@ -13,7 +13,7 @@ pool:
 
 jobs:
   - job: V8JsiBuild
-    timeoutInMinutes: 180
+    timeoutInMinutes: 210
     displayName: Build v8jsi.dll
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.ado/pull-request.yml
```diff
@@ -13,7 +13,7 @@ pool:
 
 jobs:
   - job: V8JsiBuild
-    timeoutInMinutes: 180
+    timeoutInMinutes: 210
```
NikoAri, line 16: >210 [](http://example.com/codeflow?start=22&length=3)

That's like 4 hours, isn't it a bit long?

NikoAri (approved):
:shipit:

.ado/pull-request.yml
tudorms: Yes :( But we sometimes get time-outs, especially when building UWP / ARM64 because we need to host-compile tools, and it sucks waiting 3 hours to get the timeout then re-running the slices again. Hopefully this alleviates some of those instances.

PR #72, opened 2021-08-11
Fix dll copy path

rgerd:
Now that the NuGet packages don't have `AppPlatform` folders we should remove that part of the path for the dll.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/72)


rgerd:
Fixed by 54bf58b72c84a2b8cf2ab43b5b8ded8dafafe201

PR #74, opened 2021-08-20
Update src\jsi\jsi.h to rn@4ad4426da

JunielKatarn:
See https://github.com/microsoft/react-native-windows/blob/0748834a41d71a5e3f70dfc83d0a005a83b4e8a2/vnext/package.json#L75

Updates src\jsi\jsi.h to React Native 4ad4426daf51903574f81aff1b773645a16831e0.

## Notes
- Must be backported to `0.64-stable`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/74)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.24",
+    "version": "0.65.1",
     "v8ref": "refs/branch-heads/9.3"
 }
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.23",
+    "version": "0.64.24",
     "v8ref": "refs/branch-heads/9.3"
 }
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -909,6 +909,7 @@ void V8Runtime::ReportException(v8::TryCatch *try_catch) {
       auto err = jsi::JSError(*this, ex_messages);
       err.value().getObject(*this).setProperty(
           *this, "stack", facebook::jsi::String::createFromUtf8(*this, sstr.str()));
+      err.setStack(sstr.str());
       throw err;
     } else {
       // If we're already in stack overflow, calling the Error constructor pushes it overboard
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -617,6 +617,10 @@ jsi::Value V8Runtime::evaluateJavaScript(
   return result;
 }
 
+bool V8Runtime::drainMicrotasks(int /*maxMicrotasksHint*/) {
+  return true;
+}
+
 // The callback that is invoked by v8 whenever the JavaScript 'print'
 // function is called.  Prints its arguments on stdout separated by
 // spaces and ending with a newline.
@@ -905,7 +909,6 @@ void V8Runtime::ReportException(v8::TryCatch *try_catch) {
       auto err = jsi::JSError(*this, ex_messages);
       err.value().getObject(*this).setProperty(
           *this, "stack", facebook::jsi::String::createFromUtf8(*this, sstr.str()));
-      err.setStack(sstr.str());
       throw err;
     } else {
       // If we're already in stack overflow, calling the Error constructor pushes it overboard
--- a/src/jsi/decorator.h
+++ b/src/jsi/decorator.h
@@ -126,6 +126,9 @@ class RuntimeDecorator : public Base, private jsi::Instrumentation {
       const std::shared_ptr<const PreparedJavaScript>& js) override {
     return plain().evaluatePreparedJavaScript(js);
   }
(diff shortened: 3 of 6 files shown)


config.json
```diff
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.23",
+    "version": "0.64.24",
```
NickGerleman, line 2: FYI the JSI change landed in 0.65, so RN 0.64 expects the v-table without drainMicroTasks

src/jsi/jsi.h
```diff
@@ -185,6 +185,35 @@ class Runtime {
...
+  /// Note that error propagation is only a concern if a host needs to implement
+  /// `queueMicrotask`, a recent API that allows enqueueing aribitary functions
+  /// (hence may throw) as microtasks. Exceptions from ECMA-262 Promise Jobs are
+  /// handled internally to VMs and are never propagrated to hosts.
+  ///
+  /// This API offers some queue management to hosts at its best effort due to
+  /// different behaviors and limitations imposed by different VMs and APIs. By
+  /// the time this is written, An implementation may swallow exceptions (JSC),
+  /// may not pause (V8), and may not support bounded executions.
+  virtual bool drainMicrotasks(int maxMicrotasksHint = -1) = 0;
```
NickGerleman, line 215: Does this need an implementation?

Iirc the JSC and Chakra stubs just return false, but I think @mganandraj has context into the real implementation.

config.json
JunielKatarn: I'm not sure how v8/JSI versions are handled. @tudorms, should we bump this version to `0.65.x` already?

src/jsi/jsi.h
JunielKatarn: I'll hold for @mganandraj's reply.
Right now, we're trying to solve an ABI issue when using RNW+JSI in the `main` branch.

NickGerleman: Yeah I'd just stub it for now then. That's what we did for Chakra a while back. Let me forward you a thread related to the ABI change.

config.json
tudorms: First we need to create a new branch, 0.64-stable, with the current state, then switch master to 0.65.1 along with the JSI changes.

JunielKatarn: I have created branch `0.64-stable` based on the current `master`. I'll update the version on this PR accordingly.

NikoAri (approved):
:shipit:

tudorms (approved):

src/V8JsiRuntime.cpp
```diff
@@ -1,1621 +1,1624 @@
...
-        TraceLoggingString(ex_messages.c_str(), "ex_messages"),
-        TraceLoggingString(sstr.str().c_str(), "sstr"));
-
-    // V8 doesn't actually capture the current callstack (as we're outside of scope when this gets called)
-    // See also https://v8.dev/docs/stack-trace-api
-    if (ex_messages.find("Maximum call stack size exceeded") == std::string::npos) {
-      auto err = jsi::JSError(*this, ex_messages);
-      err.value().getObject(*this).setProperty(
-          *this, "stack", facebook::jsi::String::createFromUtf8(*this, sstr.str()));
-      err.setStack(sstr.str());
```
JunielKatarn, line 908: @tudorms this method was removed from `JSError` without any apparent replacement.

Is it safe to assume we can just leave it out?
Maybe the stack is implicitly set now?

tudorms: I think we should just add that in; I added it a while back manually but we never backported to Facebok. It's needed because it's the only way to preserve the callstack in V8 (I guess for Hermes the message includes the stack)

JunielKatarn: Done.

src/jsi/decorator.h
```diff
@@ -126,6 +126,9 @@ class RuntimeDecorator : public Base, private jsi::Instrumentation {
       const std::shared_ptr<const PreparedJavaScript>& js) override {
     return plain().evaluatePreparedJavaScript(js);
   }
+  bool drainMicrotasks(int /*maxMicrotasksHint*/) {
+    return true;
```
NickGerleman, line 130: If I recall the stub returns false on other platforms. Any reason why this returns true?

NickGerleman: Actually this I think is normally just part of the JSI package and forwards bits along. See https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/ReactCommon/jsi/jsi/decorator.h

NickGerleman: My memory for the JSC one was wrong. That does return true as well 👍.

PR #90, opened 2021-10-07
Android pipeline setup

shivenmian:
Adding in the ADO pipeline and nuspec for the Android build, do not merge

### DIFF
--- a/android/ReactNative.V8Jsi.Android.nuspec
+++ b/android/ReactNative.V8Jsi.Android.nuspec
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>$Id$</id>
+    <version>$Version$</version>
+    <!-- VERSION_DETAILS => [Android] V8 version: $v8Version; Git revision: $gitRevision -->
+    <description>Contains an Android implementation of the V8 JSI wrapper for React Native ($VersionDetails$)</description>
+    <authors>Facebook, Google, Microsoft</authors>
+    <projectUrl>https://github.com/microsoft/v8-jsi/tree/master/android</projectUrl>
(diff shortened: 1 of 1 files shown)


android/localbuild.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# defining params
+
+BUILD_PLATFORM="x64" # default
+BUILD_FLAVOR="debug"
+SOURCE_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+SOURCES_PATH="$SOURCE_DIR" # this won't be supplied by the user; should be android folder
+OUTPUT_PATH="$SOURCE_DIR/out"
+
+while [ "${1:-}" != "" ]; do
```
NikoAri, line 18: >${1:-} [](http://example.com/codeflow?start=9&length=6)

(nit) These shell languages are hardly readable. Can we convert this to PowerShell, JavaScript, Perl, or any other non-shell language?

.gitignore
```diff
@@ -1,5 +1,8 @@
 build
-out
 .vscode
 build.log
-src/version_gen.rc
\ No newline at end of file
+src/version_gen.rc
+
+# outputs
+out*
```
NikoAri, line 7: >* [](http://example.com/codeflow?start=3&length=1)

(nit) do we need this *? Seems like an overkill

shivenmian: Fixed

android/localbuild.sh
JunielKatarn: Unless this project is meant to run only on Linux hosts, converting the build logic to PowerShell (which also runs on Linux) gives us a more uniform code base.

In my opinion, bash-only complex logic should only be written for UNIX-only projects.

shivenmian:
+1. Adding to what @tudorms said, currently we are using the Android build mostly for [react-native-android](https://github.com/microsoft/react-native-macos) and other RN Android projects, which are mostly Linux based right now. Besides, the script that the Android build uses to install v8 dependencies can only run on Linux.

PR #96, opened 2021-11-20
Update to V8 9.7

tudorms:
- Update to V8 9.7
- Manually clear the list of HostObjectLifetimeTracker (in case there are leaks at shutdown, to prevent crashing)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/96)

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -564,6 +564,8 @@ V8Runtime::~V8Runtime() {
     hostObjectLifetimeTracker->ResetHostObject(false /*isGC*/);
   }
 
+  host_object_lifetime_tracker_list_.clear();
+
 #if defined(_WIN32) && defined(V8JSI_ENABLE_INSPECTOR)
   if (inspector_agent_){
     inspector_agent_.reset();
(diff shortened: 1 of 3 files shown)


src/V8JsiRuntime.cpp
```diff
@@ -1,1628 +1,1630 @@
...
-  if (inspector_agent) {
-    inspector_agent->start();
-  }
-}
-
-void openInspectors_toberemoved() { inspector::Agent::startAll(); }
-#endif
-
-} // namespace v8runtime
+// Copyright (c) Microsoft Corporation.
```
JunielKatarn, line 1: What changed?

Line endings?

JunielKatarn: Diffed manually.

Apparently, this version only introduces this line:
```
host_object_lifetime_tracker_list_.clear();
```

JunielKatarn (approved):

src/V8JsiRuntime.cpp
tudorms: Yes, that's the only difference. GitHub is being weird, but CodeFlow will load the diff correctly.

PR #98, opened 2021-11-24
[0.63] Update to V8 9.7

tudorms:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/98)

### DIFF
--- a/src/inspector/inspector_socket_server.cpp
+++ b/src/inspector/inspector_socket_server.cpp
@@ -338,7 +338,8 @@ bool InspectorSocketServer::HasTargets() {
 
 bool InspectorSocketServer::Start() {
   state_ = ServerState::kRunning;
-  std::thread([this]() {
+  std::unique_lock<std::mutex> tcp_server_lock(mutex_tcp_server_);
+  std::thread([this, lock = std::move(tcp_server_lock)]() {
     tcp_server_ = std::make_shared<tcp_server>(
         port_, InspectorSocketServer::SocketConnectedCallback, this);
     tcp_server_->run();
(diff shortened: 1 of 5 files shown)


Khalef1:
Approving, but note that Office has moved to 0.64.

Khalef1 (approved):

NikoAri (approved):
:shipit:

src/inspector/inspector_socket_server.cpp
```diff
@@ -338,7 +338,8 @@ bool InspectorSocketServer::HasTargets() {
 
 bool InspectorSocketServer::Start() {
   state_ = ServerState::kRunning;
-  std::thread([this]() {
+  std::unique_lock<std::mutex> tcp_server_lock(mutex_tcp_server_);
+  std::thread([this, lock = std::move(tcp_server_lock)]() {
```
NikoAri, line 342: >this [](http://example.com/codeflow?start=15&length=4)

(nit) are we ok to pass raw pointer here?

NikoAri, line 342: >lock [](http://example.com/codeflow?start=21&length=4)

So this is locked even before thread has a chance to execute. Should this be scoped down to start of execution?

NikoAri: If such extended lock is indeed an expected behavior, would you mind adding brief comment explaining it?

tudorms: It feels safer to lock before; don't know what would happen if we get through Start, then Stop, then the thread gets scheduled now touching deleted memory.

FWIW, in real-life this won't really matter; it's an issue in unit tests that quickly bring up and tear down runtimes (see 5610714 in Office).

NikoAri: I see. If it's not too difficult, please make a comment in code about it.

PR #99, opened 2021-11-24
Protect tcp_server access

tudorms:
If the runtime is quickly created and torn down with inspector enabled we could end up trying to destroy tcp_server before it finishes creating.

This change protects its creation / destruction with a mutex.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/99)

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -704,7 +704,7 @@ jsi::Value V8Runtime::ExecuteString(const v8::Local<v8::String> &source, const s
   std::shared_ptr<const jsi::Buffer> cache;
   if (args_.preparedScriptStore) {
     jsi::ScriptSignature scriptSignature = {sourceURL, 1};
-    jsi::JSRuntimeSignature runtimeSignature = {"V8", 76};
+    jsi::JSRuntimeSignature runtimeSignature = {"V8", V8_MAJOR_VERSION * 10000 + V8_MINOR_VERSION * 1000 + V8_BUILD_NUMBER};
     cache = args_.preparedScriptStore->tryGetPreparedScript(scriptSignature, runtimeSignature, "perf");
   }
 
@@ -740,7 +740,7 @@ jsi::Value V8Runtime::ExecuteString(const v8::Local<v8::String> &source, const s
         v8::ScriptCompiler::CachedData *codeCache = v8::ScriptCompiler::CreateCodeCache(script->GetUnboundScript());
 
         jsi::ScriptSignature scriptSignature = {sourceURL, 1};
-        jsi::JSRuntimeSignature runtimeSignature = {"V8", 76};
+        jsi::JSRuntimeSignature runtimeSignature = {"V8", V8_MAJOR_VERSION * 10000 + V8_MINOR_VERSION * 1000 + V8_BUILD_NUMBER};
 
         args_.preparedScriptStore->persistPreparedScript(
             std::make_shared<ByteArrayBuffer>(codeCache->data, codeCache->length),
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -701,10 +701,12 @@ jsi::Value V8Runtime::ExecuteString(const v8::Local<v8::String> &source, const s
   v8::ScriptCompiler::CompileOptions options = v8::ScriptCompiler::CompileOptions::kNoCompileOptions;
   v8::ScriptCompiler::CachedData *cached_data = nullptr;
 
+  jsi::JSRuntimeVersion_t runtimeVersion = V8_MAJOR_VERSION * 10000 + V8_MINOR_VERSION * 1000 + V8_BUILD_NUMBER;
+
   std::shared_ptr<const jsi::Buffer> cache;
   if (args_.preparedScriptStore) {
     jsi::ScriptSignature scriptSignature = {sourceURL, 1};
-    jsi::JSRuntimeSignature runtimeSignature = {"V8", V8_MAJOR_VERSION * 10000 + V8_MINOR_VERSION * 1000 + V8_BUILD_NUMBER};
+    jsi::JSRuntimeSignature runtimeSignature = {"V8", runtimeVersion};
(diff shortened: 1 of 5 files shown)


src/inspector/inspector_socket_server.cpp
```diff
@@ -338,7 +338,11 @@ bool InspectorSocketServer::HasTargets() {
 
 bool InspectorSocketServer::Start() {
   state_ = ServerState::kRunning;
-  std::thread([this]() {
+
+  // It feels safer to lock before the thread jump; if we get through Start, then Stop, then the thread gets scheduled now touching deleted memory that's not good either.
+  // In real-life this won't really matter; it's only an issue in unit tests that quickly bring up and tear down runtimes.
+  std::unique_lock<std::mutex> tcp_server_lock(mutex_tcp_server_);
+  std::thread([this, lock = std::move(tcp_server_lock)]() {
     tcp_server_ = std::make_shared<tcp_server>(
```
vmoroz, line 346: >tcp_server_ [](http://example.com/codeflow?start=4&length=11)

It seems that we still have a race condition: the thread can start a little bit later than the Stop is executed.
It would be better to use the mutex to have any access to the tcp_server_ and state_ variables.
Also, we probably need a condition variable to notify code in the Stop method that the thread is finished.

tudorms: We lock in Start, then move the unique_lock to the thread, where it unlocks when the async callback is finished; Start / Stop must be called from the same thread. I think the change takes care of race access conditions on tcp_server_ but I agree we need to take another look at the implementation top-down.

vmoroz: Locks are per thread. We must not move them to another thread. We just need to create a new lock inside of the thread when we access the tcp_server_ variable. Inside of the lambda we need to create a local variable for it to make sure that we can unlock the lock before we call run.

vmoroz: If we could move the lock to another thread it would never be released because it would be released only when the lambda is destroyed and the the Stop would never able to acquire the mutex. The issue is not with locks, but with mutexes which are typically bound to a thread when locked, i.e. if a thread acquires a mutex, then other threads cannot acquire it and must wait.

tudorms: Thank you, that makes sense.

src/V8JsiRuntime.cpp
```diff
@@ -740,7 +740,7 @@ jsi::Value V8Runtime::ExecuteString(const v8::Local<v8::String> &source, const s
         v8::ScriptCompiler::CachedData *codeCache = v8::ScriptCompiler::CreateCodeCache(script->GetUnboundScript());
 
         jsi::ScriptSignature scriptSignature = {sourceURL, 1};
-        jsi::JSRuntimeSignature runtimeSignature = {"V8", 76};
+        jsi::JSRuntimeSignature runtimeSignature = {"V8", V8_MAJOR_VERSION * 10000 + V8_MINOR_VERSION * 1000 + V8_BUILD_NUMBER};
```
NikoAri, line 743: >V8_MAJOR_VERSION * 10000 [](http://example.com/codeflow?start=58&length=24)

(nit) can you create a function returning this signature to avoid duplicated calculation

src/inspector/inspector_socket_server.cpp
```diff
@@ -338,9 +338,17 @@ bool InspectorSocketServer::HasTargets() {
 
 bool InspectorSocketServer::Start() {
   state_ = ServerState::kRunning;
+
   std::thread([this]() {
```
NikoAri, line 342: >this [](http://example.com/codeflow?start=15&length=4)

So, what component ensures that this object is alive by the time lambda executes? Shouldn't it be a weak pointer?

NikoAri (approved):
:shipit:

src/inspector/inspector_socket_server.h
```diff
@@ -81,7 +81,9 @@ class InspectorSocketServer {
   const std::string host_;
   int port_;
 
+  std::mutex mutex_tcp_server_;
   std::shared_ptr<tcp_server> tcp_server_;
+  bool tcp_server_stopped_{false};
```
vmoroz, line 86: >tcp_server_stopped_ [](http://example.com/codeflow?start=7&length=19)

Can we instead rely on the ServerState state_ field and protect it by the mutex too? Otherwise, we are effectively defining two different fields that keep the state of the server which are not synchronized with each other.

src/inspector/inspector_socket_server.cpp
```diff
@@ -338,10 +338,18 @@ bool InspectorSocketServer::HasTargets() {
 
 bool InspectorSocketServer::Start() {
   state_ = ServerState::kRunning;
```
vmoroz, line 340: >state_ = ServerState::kRunning; [](http://example.com/codeflow?start=2&length=31)

Since on the Stop() operation we reset the delegate, I assume that we must not be able to Start the server after it was stopped. If it is correct, then we must check the state_ to ensure that we cannot change it to kRunning after it is already kStopped.

src/inspector/inspector_socket_server.h
vmoroz: Though, the keeping the separate bool tcp_server_stopped_ can be still simpler as long as we check the state_ variable in Start before we change it.

vmoroz (approved):
:shipit:

src/inspector/inspector_socket_server.cpp
tudorms: Yes, that cannot happen; added a CHECK.

src/inspector/inspector_socket_server.h
tudorms: I'll keep them separate. The object is already synchronously used and guaranteed to never Stop before Starting.

The only issue we need to address is with the tcp_server member which runs on a separate thread and in UTs we can actually call Stop before tcp_server's thread has a chance to get scheduled and start (would likely never happen in real use scenarios).

NikoAri (approved):

PR #100, opened 2021-11-30
Populate a real runtime signature for PreparedScriptStore

tudorms:
Use the V8 build number to compose a "runtime signature" for use by the script cache; this would ensure we won't try to load script caches from potentially incompatible V8 builds.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/100)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.64.28",
+    "version": "0.64.29",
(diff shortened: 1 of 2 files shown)


vmoroz (approved):
:shipit:

PR #104, opened 2021-12-10
Call globalInitializeTracing only once

tudorms:
Make sure globalInitializeTracing is only called once
Also port misc changes from master

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/104)

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -23,34 +23,36 @@ Push-Location (Join-Path $workpath "v8build\v8")
 # Generate the build system
 $gnargs = 'v8_enable_i18n_support=false is_component_build=false v8_monolithic=true v8_use_external_startup_data=false treat_warnings_as_errors=false'
 
-if ($Configuration -like "*android") {
-    $gnargs += ' use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
+# commenting this out for now since Android build is at android/build.sh
+
+# if ($Configuration -like "*android") {
+#     $gnargs += ' use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
+# }
+# else {
+if (-not ($UseLibCpp)) {
+    $gnargs += ' use_custom_libcxx=false'
+}
+
+if ($AppPlatform -eq "uwp") {
+    # the default target_winuwp_family="app" (which translates to WINAPI_FAMILY=WINAPI_FAMILY_PC_APP) blows up with too many errors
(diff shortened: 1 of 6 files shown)


scripts/build.ps1
```diff
@@ -23,34 +23,36 @@ Push-Location (Join-Path $workpath "v8build\v8")
 # Generate the build system
 $gnargs = 'v8_enable_i18n_support=false is_component_build=false v8_monolithic=true v8_use_external_startup_data=false treat_warnings_as_errors=false'
 
-if ($Configuration -like "*android") {
-    $gnargs += ' use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
+# commenting this out for now since Android build is at android/build.sh
+
+# if ($Configuration -like "*android") {
```
NikoAri, line 28: ># if  [](http://example.com/codeflow?start=0&length=5)

(nit) should all this just be removed?

scripts/fetch_code.ps1
```diff
@@ -15,13 +15,14 @@ $env:GIT_REDIRECT_STDERR = '2>&1'
 Push-Location (Join-Path $workpath "v8build")
 & fetch v8
 
-if ($Configuration -like "*android") {
-    $target_os = @'
-target_os= ['android']
-'@
+# commenting this out since the Android build is in android/build.sh
+# if ($Configuration -like "*android") {
```
NikoAri, line 19: (nit) same comment as in other file, Remove this?

src/V8JsiRuntime.cpp
```diff
@@ -446,9 +446,11 @@ void V8Runtime::createHostObjectConstructorPerContext() {
       isolate_, constructorForHostObjectTemplate->GetFunction(context_.Get(isolate_)).ToLocalChecked());
 }
 
+std::once_flag g_tracingInitialized;
```
NikoAri, line 449: >g_tracingInitialized [](http://example.com/codeflow?start=15&length=20)

(nit) should this be declared inside of ifdef _WIN32 clause?

NikoAri (approved):
:shipit:

src/V8JsiRuntime.cpp
tudorms: Thanks, good catch.

scripts/build.ps1
tudorms: I'll leave the scripts like this for now to match master.

PR #107, opened 2022-02-07
Update Node-API tests from hermes-windows repo

vmoroz:
The resent PR https://github.com/microsoft/hermes-windows/pull/61 has implemented Node-API for Hermes. In that PR we had to update the unit test suite to be compatible with both implementations: V8 and Hermes.

In this PR we have:
- brought changes to Node-API tests from the https://github.com/microsoft/hermes-windows repo;
- removed `.js.h` files that used to wrap up the `.js` files. Instead, we copy the `.js` files next to the test binary in the `BUILD.gn` and then use them while running unit tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/107)


tudorms (approved):
:shipit:

PR #111, opened 2022-02-24
Make publish pipeline compliant

dannyvv:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/111)

### DIFF
--- a/.ado/windows-build.yml
+++ b/.ado/windows-build.yml
@@ -65,13 +65,26 @@ steps:
       ignoreDirectories: 'build\v8build\depot_tools'
 
   - script: mkdir ${{parameters.outputPath}}\_manifest\$(BuildPlatform)\$(BuildConfiguration)
+    displayName: 📒 Prep Manifest
 
   - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
     displayName: 📒 Generate Manifest
     inputs:
       BuildDropPath: ${{parameters.outputPath}}
       ManifestDirPath: ${{parameters.outputPath}}\_manifest\$(BuildPlatform)\$(BuildConfiguration)
-      
+
+  # Guardian does not handle custom builds, so manually running Binskim
+  - task: BinSkim@3
+    displayName: Run Binskim Analysis
+    inputs:
+        InputType: 'Basic'
+        Function: 'analyze'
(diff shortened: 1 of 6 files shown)


.ado/windows-build.yml
```diff
@@ -65,17 +65,32 @@ steps:
...
+
   - task: PublishBuildArtifacts@1
     displayName: "Publish artifacts"
     inputs:
       artifactName: V8Jsi
       pathtoPublish: ${{parameters.outputPath}}
 
   - task: ComponentGovernanceComponentDetection@0
+    inputs:
+      ignoreDirectories: 'android/build/v8/third_party/catapult'
```
tudorms, line 96: >ignoreDirectories: 'android/build/v8/third_party/catapult' [](http://example.com/codeflow?start=6&length=58)

This directory should not exist for Windows builds.

scripts/build.ps1
```diff
@@ -118,10 +119,10 @@ if (!$PSVersionTable.Platform -or $IsWindows) {
         Copy-Item "$buildoutput\v8jsi.dll.pdb" -Destination "$OutputPath\lib\$AppPlatform\$Configuration\$Platform"
     }
 
-    # Debugging extension
-    if ($AppPlatform -ne "uwp") {
-        Copy-Item "$buildoutput\v8windbg.dll" -Destination "$OutputPath\lib\$AppPlatform\$Configuration\$Platform"
-    }
+    # Skip Debugging extension
```
tudorms, line 122: >Skip Debugging extension [](http://example.com/codeflow?start=6&length=24)

We do want to publish this into the NuGet as it can be manually retrieved for use in debugging (it needs to version match v8jsi.dll). Is there a way to skip scanning it instead of removing it altogether?

dannyvv: From the docs there is no way to exclude it from binskim...
@tudorms : The only option to get it passing is to rename run tool and name back.. But that is agsinst the spirit of compliance. I would not recommend it... We should either get it compliant or not ship it...

vmoroz:
Please add the PR description.

NikoAri (approved):
:shipit:

scripts/build.ps1
dannyvv: Resolved with Tudor

tudorms (approved):
:shipit:

PR #122, opened 2022-04-18
Fix Node-API calling conventions

vmoroz:
### The issue

When we tried to adopt the Node-API based code in [ReactNative for Windows](https://github.com/microsoft/react-native-windows) we have got runtime crashes in x86 Release mode.

The issue is that some function pointers in Node-API did not have explicit calling conventions and since `react-native-win32.dll` produced by [ReactNative for Windows](https://github.com/microsoft/react-native-windows) uses `__stdcall` calling conventions by default while most of Node-API is `__cdecl`, we have got a mismatch in calling conventions for all function pointers passed to Node-API. The Node-API is expecting `__cdecl` function pointer and given a `__stdcall` function pointer instead. It causes a runtime crash.

### The solution

We set the `__cdecl` explicitly for function pointers in `js_native_api_types.h` and `js_native_ext_api.h`.

We also remove implementation of JSI for Node-API instead of fixing it because it is not being used.
Instead, we use code defined in the [ReactNative for Windows](https://github.com/microsoft/react-native-windows) repo:

- [NodeApiJsiRuntime.h](https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Cxx/JSI/NodeApiJsiRuntime.h)
- [NodeApiJsiRuntime.cpp](https://github.com/microsoft/react-native-windows/blob/main/vnext/Microsoft.ReactNative.Cxx/JSI/NodeApiJsiRuntime.cpp)

This implementation will be updated after we release a new version of `v8jsi.dll` Nuget package.


tudorms:
It was nice to have NodeApiJsiRuntime.h/.cpp alongside the napi implementation; we could test it at the source, and make the package consumable stand-alone (without RNW). But if we add napi+jsi tests in RNW that should provide some coverage as well.

tudorms (approved):
:shipit:

NikoAri (approved):

src/public/js_native_ext_api.h
```diff
@@ -28,13 +28,13 @@ typedef struct napi_ext_env_scope__ *napi_ext_env_scope;
 typedef struct napi_ext_ref__ *napi_ext_ref;
 
 // A callback to return buffer synchronously
-typedef void (*napi_ext_buffer_callback)(napi_env env, uint8_t const *buffer, size_t buffer_length, void *buffer_hint);
+typedef void (__cdecl *napi_ext_buffer_callback)(napi_env env, uint8_t const *buffer, size_t buffer_length, void *buffer_hint);
```
NikoAri, line 31: >__cdecl [](http://example.com/codeflow?start=14&length=7)

Can we add some tests that verify that we don't lose this signature over time?

vmoroz:
> It was nice to have NodeApiJsiRuntime.h/.cpp alongside the napi implementation; we could test it at the source, and make the package consumable stand-alone (without RNW). But if we add napi+jsi tests in RNW that should provide some coverage as well.

I see your point and it was not an easy decision. Currently we have three copies of this code: here, hermes-windows, and RNW. I want to have a single copy instead. Having a full repo just for these few lines seems to be too much. My current thoughts to have all this code inside of RNW, but like you said it makes it difficult to use v8jsi.dll package on its own.
I am open to other ideas.

src/public/js_native_ext_api.h
vmoroz: I cannot think about any such test. Any ideas?

PR #123, opened 2022-04-19
Build improvements

tudorms:
- Refactor / clean up build scripts (should improve build speed by 5%+)
- Add experimental support for building on/for Android, Linux and Mac with the same PowerShell build script

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/123)

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -25,10 +25,10 @@ Push-Location (Join-Path $workpath "v8build\v8")
 $gnargs = 'v8_enable_i18n_support=false is_component_build=false v8_monolithic=true v8_use_external_startup_data=false treat_warnings_as_errors=false'
 
 if ($Configuration -like "*android") {
-    $gnargs += ' use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
+    $gnargs += ' v8jsi_enable_napi=false use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
 }
 elseif ($Configuration -like "*linux") {
-    $gnargs += ' use_goma=false target_os=\"linux\" target_cpu=\"' + $Platform + '\"'
+    $gnargs += ' v8jsi_enable_napi=false use_goma=false target_os=\"linux\" target_cpu=\"' + $Platform + '\"'
 }
 else {
 if (-not ($UseLibCpp)) {
@@ -61,7 +61,11 @@ if ($Platform -like "?64") {
 # }
 
 if ($Configuration -like "*ebug*") {
-    $gnargs += ' enable_iterator_debugging=true is_debug=true'
+    $gnargs += ' is_debug=true'
+
+    if (($Configuration -notlike "*android") -and ($Configuration -notlike "*linux")) {
+        $gnargs += ' enable_iterator_debugging=true'
+    }
 }
 else {
     $gnargs += ' enable_iterator_debugging=false is_debug=false'
@@ -165,6 +169,7 @@ Copy-Item "$jsigitpath\jsi\instrumentation.h" -Destination "$OutputPath\build\na
 
 # Miscellaneous
 Copy-Item "$SourcesPath\ReactNative.V8Jsi.Windows.targets" -Destination "$OutputPath\build\native\"
+Copy-Item "$SourcesPath\ReactNative.V8Jsi.Windows.nuspec" -Destination $OutputPath
 
 Copy-Item "$SourcesPath\config.json"    -Destination "$OutputPath\"
 Copy-Item "$SourcesPath\LICENSE"        -Destination "$OutputPath\license\"
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -24,32 +24,26 @@ Push-Location (Join-Path $workpath "v8build\v8")
 # Generate the build system
 $gnargs = 'v8_enable_i18n_support=false is_component_build=false v8_monolithic=true v8_use_external_startup_data=false treat_warnings_as_errors=false'
 
-if ($AppPlatform -eq "android") {
-    $gnargs += ' v8jsi_enable_napi=false use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
-}
-elseif ($AppPlatform -eq "linux") {
-    $gnargs += ' v8jsi_enable_napi=false use_goma=false target_os=\"linux\" target_cpu=\"' + $Platform + '\"'
-}
-else {
-if (-not ($UseLibCpp)) {
-    $gnargs += ' use_custom_libcxx=false'
-}
+if ("android linux mac".contains($AppPlatform)) {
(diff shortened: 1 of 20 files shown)


src/BUILD.gn
```diff
@@ -14,24 +15,29 @@ target("shared_library", "v8jsi") {
...
-    "public/js_native_ext_api.h",
-    "public/NapiJsiRuntime.cpp",
     "public/ScriptStore.h",
     "public/V8JsiRuntime.h",
     "V8JsiRuntime_impl.h",
     "V8JsiRuntime.cpp",
     "IsolateData.h",
   ]
 
+  if (v8jsi_enable_napi) {
```
dannyvv, line 25: should we use the full name?
```suggestion
  if (v8jsi_enable_node_api) {
```

scripts/build.ps1
```diff
@@ -23,41 +24,41 @@ Push-Location (Join-Path $workpath "v8build\v8")
...
-
-# if ($Configuration -like "*android") {
-#     $gnargs += ' use_goma=false target_os=\"android\" target_cpu=\"' + $Platform + '\"'
-# }
-# else {
-if (-not ($UseLibCpp)) {
-    $gnargs += ' use_custom_libcxx=false'
-}
+if ("android linux mac".contains($AppPlatform)) {
+    $gnargs += ' v8jsi_enable_napi=false use_goma=false target_os=\"' + $AppPlatform + '\"'
```
mganandraj, line 28: >v8jsi_enable_napi=false [](http://example.com/codeflow?start=17&length=23)

Curious, are there any big issues with enabling current napi implementation on non-windows ?

scripts/build.ps1
```diff
@@ -23,41 +24,41 @@ Push-Location (Join-Path $workpath "v8build\v8")
...
 
 $gnargs += ' target_cpu=\"' + $Platform + '\"'
 
-if ($UseClang) {
-    #TODO (#2): we need to figure out how to actually build DEBUG with clang-cl (won't work today due to STL iterator issues)
-    $gnargs += ' is_clang=true'
-}
-else {
-    $gnargs += ' is_clang=false'
+if (!"android linux mac".contains($AppPlatform)) {
```
mganandraj, line 42: >"android linux mac".contains($AppPlatform) [](http://example.com/codeflow?start=5&length=42)

nit: This check is repeated many times .. could have a single flag, say 'isWin' defined on top ..

scripts/build.ps1
```diff
@@ -71,23 +72,35 @@ if (!$?) {
...
+& ninja -v -j $numberOfThreads -C $buildoutput v8jsi $ninjaExtraTargets | Tee-Object -FilePath "$SourcesPath\build.log"
 if (!$?) {
     Write-Host "Build failure, check logs for details"
     exit 1
 }
 
 Pop-Location
 
-if (!(Test-Path -Path "$buildoutput\v8jsi.dll") -and !(Test-Path -Path "$buildoutput\libv8jsi.so")) {
+if (!(Test-Path -Path "$buildoutput\v8jsi.dll") -and !(Test-Path -Path "$buildoutput\libv8jsi.so") -and !(Test-Path -Path "$buildoutput\libv8jsi.dylib")) {
```
mganandraj, line 103: >libv8jsi.so [](http://example.com/codeflow?start=85&length=11)

Is it tested ? Is it linking against static CRT ?

mganandraj (approved):
:shipit:

scripts/build.ps1
```diff
@@ -71,23 +72,35 @@ if (!$?) {
 }
 
 # We'll use 2x the number of cores for parallel execution
-$numberOfThreads = [int]((Get-CimInstance Win32_ComputerSystem).NumberOfLogicalProcessors) * 2
+$numberOfThreads = 2
+if ($PSVersionTable.Platform -and !$IsWindows) {
+    if ($IsMacOS) {
+        $numberOfThreads = [int](Invoke-Command -ScriptBlock { sysctl -n hw.physicalcpu }) * 2
```
vmoroz, line 78: >     [](http://example.com/codeflow?start=94&length=4)

not: please remove trailing spaces

scripts/build.ps1
tudorms: Not sure about the static CRT.

I ran jsitests on Linux against libv8jsi.so and they mostly all run (except for the ones involving exceptions, for some reason the exception type isn't being preserved - I didn't investigate that too deep).

src/BUILD.gn
```diff
@@ -81,8 +90,9 @@ target("shared_library", "v8jsi") {
     "//:v8_monolith"
   ]
 
+  defines = [ "BUILDING_V8_SHARED" ]
```
vmoroz, line 93: >BUILDING_V8_SHARED [](http://example.com/codeflow?start=15&length=18)

This flag exposes all V8 API as public functions. As a result the public API surface for the DLL is huge.
It would be better if we define a separate definition for JSI API only.

src/makev8jsi.lst
```diff
@@ -0,0 +1,6 @@
+{
```
vmoroz, line 1: >  [](http://example.com/codeflow?start=1&length=1)

nit: please remove trailing spaces in the end of all lines in this file.

src/V8JsiRuntime.cpp
```diff
@@ -1599,19 +1605,21 @@ napi_ext_ref NapiUniqueString::GetRef() const noexcept {
...
   string_ref_ = ref;
 }
+#endif
 
 //=============================================================================
 // Make V8 JSI runtime
 //=============================================================================
 
-std::unique_ptr<jsi::Runtime> makeV8Runtime(V8RuntimeArgs &&args) {
+#ifdef _WIN32
```
vmoroz, line 1614: >ifdef [](http://example.com/codeflow?start=1&length=5)

Would it be better to define a macro V8JSI_EXPORT inside of V8JsiRuntime.h?

scripts/build.ps1
tudorms: I had more errors when I first started, but just tried it now and it doesn't look so bad. I think I'll just fix those and just get rid of this build variable.

vmoroz:
> Copy-Item "$jsigitpath\public\NapiJsiRuntime.h" -Destination "$OutputPath\build\native\include\"

Should we avoid copying Node-API files if we do not build them?

---
Refers to: scripts/build.ps1:152 in b3ebb5f. [](commit_id = b3ebb5f7077b16698403aad79301ce771d65631e, deletion_comment = False)

vmoroz (approved):
:shipit:

scripts/build.ps1
mganandraj: We need v8jsi linked against dynamic CRT to be consumed in RN .. so that RN and v8jsi works with the same types ..

PR #124, opened 2022-05-02
Implement hash for script versioning

tudorms:
Since we hardcode the script version to "1" we could get into a state where we're using bytecode / script cache from an older bundle version. This change implements a simple hashing algorithm to version the scripts.

Bonus knowing the script is Ascii lets us skip a copy, and some generic cleanup.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/124)

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -6,6 +6,7 @@
 #include "v8.h"
 
 #include "IsolateData.h"
+#include "MurmurHash.h"
 #include "napi/js_native_api_v8.h"
 #include "public/ScriptStore.h"
 #include "public/js_native_api.h"
@@ -603,18 +604,8 @@ jsi::Value V8Runtime::evaluateJavaScript(
   constexpr std::uint64_t P1 {7};
   constexpr std::uint64_t P2 {31};
 
-  std::uint64_t hash {P1};
-  bool isAscii {true};
-
-  const char* p = reinterpret_cast<const char *>(buffer->data());
-  for (size_t i=0; i < buffer->size(); p++, i++) {
-    hash = hash * P2 + *p;
-
-    if (static_cast<unsigned char>(*p) > 127)
-    {
-      isAscii = false;
-    }
-  }
+  std::uint64_t hash {0};
+  bool isAscii = murmurhash(buffer->data(), buffer->size(), hash);
 
   v8::Local<v8::String> sourceV8String;
   if (isAscii) {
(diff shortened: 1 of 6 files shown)


src/V8JsiRuntime.cpp
```diff
@@ -606,24 +599,42 @@ jsi::Value V8Runtime::evaluateJavaScript(
...
+  // Basic hashing of the buffer for cache versioning
+  constexpr std::uint64_t P1 {7};
+  constexpr std::uint64_t P2 {31};
+
+  std::uint64_t hash {P1};
+  bool isAscii {true};
+
+  const char* p = reinterpret_cast<const char *>(buffer->data());
+  for (size_t i=0; i < buffer->size(); p++, i++) {
+    hash = hash * P2 + *p;
```
mganandraj, line 611: Tudor, Is this a well-know algorithm ? If i understand correctly, it loops through each chars multiplying the accumulator with 31 (left shift 5 times and subtract once ?) and add current char to accumulator .. I see it might work as the result might depend on all the previous chars .. One possible drawback would be that there is 'n' multiplications here .. 

Did you consider alternatives, such as this one: [https://en.wikipedia.org/wiki/Jenkins_hash_function#one_at_a_time](https://en.wikipedia.org/wiki/Jenkins_hash_function#one_at_a_time)

I think, shifting, xoring and adding is a lot more performant than multiplications ..

tudorms: No, this is just my go-to default hashing algorithm with primes (non-crypto) when I just need uniqueness. I'm open to whatever algorithm makes sense as long as it's small and unobtrusive (we don't need crypto / signing, just uniqueness).

tudorms: FWIW, just benchmarked the silly implementation against the Jenkins, and Jenkins is 2.5x slower.

NikoAri, line 611: >hash * P2 + *p [](http://example.com/codeflow?start=11&length=14)

Would be good to add some link to description of this algorithm

NikoAri (approved):
:shipit:

src/V8JsiRuntime.cpp
tudorms: I found hashing algorithms that are even faster (Murmur, Meow, Spooky) but the source code is huge.

tudorms: This looks like a black hole of research, I think I'll just use the std::hash<> implementation and don't bother optimizing.

vmoroz: May I suggest using the std::hash algorithm instead - it is basically the FNV-1 algorithm which is known to be quite fast and has a good distribution.

vmoroz (approved):
:shipit:

PR #130, opened 2022-05-24
[0.65] Implement PreparedScriptStore for V8 Node-API

vmoroz:
In this PR we add new Node-API extension capability to run JavaScript functions that can use the prepared script store. 
The solution consists of the following parts:

- new function `napi_ext_run_script_buffer` that can run JavaScript from the provided buffer.
- the `napi_ext_env_settings` struct is extended with a new field `script_cache` of type `napi_ext_script_cache`.
- new `napi_ext_script_cache` struct that represents the prepared script store.
- new `napi_ext_cached_script_metadata` struct with JavaScript metadata that used for validating byte code cache.
- new `napi_ext_buffer` struct to represent script text or binary buffer.
- new `napi_ext_native_data` struct as a base type to control lifetime of an external object.

The implementation wraps up the `napi_ext_script_cache` into `NodeApiPreparedScriptStore` class that implements `facebook::jsi::PreparedScriptStore` interface, which then passed to the internal `V8Runtime` instance.
Then, the `napi_ext_run_script_buffer` calls the `V8Runtime::loadJavaScript` and `V8Runtime::ExecuteString` methods to evaluate the script that internally relies on the `NodeApiPreparedScriptStore`.

The `V8Runtime::ExecuteString` is modified to return `v8::Local<v8::Value>` instead of `facebook::jsi::Value` to use the result for the Node-API `napi_value`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/130)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.65.14",
+    "version": "0.65.15",
     "v8ref": "refs/branch-heads/10.0"
 }
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -628,7 +628,7 @@ jsi::Value V8Runtime::evaluateJavaScript(
 
   std::uint64_t hash{0};
   v8::Local<v8::String> sourceV8String = loadJavaScript(buffer, hash);
-  jsi::Value result = ExecuteString(sourceV8String, sourceURL, hash);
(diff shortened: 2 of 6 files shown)


tudorms (approved):
:shipit:

PR #147, opened 2022-10-10
Multi threaded support

tudorms:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/147)

### DIFF
--- a/src/V8JsiRuntime.cpp
+++ b/src/V8JsiRuntime.cpp
@@ -20,14 +20,6 @@
 #include <mutex>
 #include <sstream>
 
-#ifndef _ISOLATE_CONTEXT_ENTER
-#define _ISOLATE_CONTEXT_ENTER                      \
-  v8::Isolate *isolate = v8::Isolate::GetCurrent(); \
-  v8::Isolate::Scope isolate_scope(isolate);        \
-  v8::HandleScope handle_scope(isolate);            \
-  v8::Context::Scope context_scope(context_.Get(isolate));
-#endif
-
 using namespace facebook;
 
 namespace v8runtime {
@@ -426,7 +418,7 @@ v8::Isolate *V8Runtime::CreateNewIsolate() {
 
 void V8Runtime::createHostObjectConstructorPerContext() {
   // Create and keep the constructor for creating Host objects.
-  v8::Local<v8::FunctionTemplate> constructorForHostObjectTemplate = v8::FunctionTemplate::New(isolate_);
+  v8::Local<v8::FunctionTemplate> constructorForHostObjectTemplate = v8::FunctionTemplate::New(GetIsolate());
   v8::Local<v8::ObjectTemplate> hostObjectTemplate = constructorForHostObjectTemplate->InstanceTemplate();
   hostObjectTemplate->SetHandler(v8::NamedPropertyHandlerConfiguration(
       HostObjectProxy::Get, HostObjectProxy::Set, nullptr, nullptr, HostObjectProxy::Enumerator));
@@ -437,7 +429,7 @@ void V8Runtime::createHostObjectConstructorPerContext() {
   hostObjectTemplate->SetIndexedPropertyHandler(HostObjectProxy::GetIndexed, HostObjectProxy::SetIndexed);
   hostObjectTemplate->SetInternalFieldCount(1);
   host_object_constructor_.Reset(
-      isolate_, constructorForHostObjectTemplate->GetFunction(context_.Get(isolate_)).ToLocalChecked());
+      GetIsolate(), constructorForHostObjectTemplate->GetFunction(GetContextLocal()).ToLocalChecked());
 }
 
 #ifdef _WIN32
@@ -536,7 +528,7 @@ V8Runtime::V8Runtime(V8RuntimeArgs &&args) : args_(std::move(args)) {
 
   const char *context_name =
       args_.debuggerRuntimeName.empty() ? "JSIRuntime context" : args_.debuggerRuntimeName.c_str();
-  inspector_agent_->addContext(context_.Get(GetIsolate()), context_name);
+  inspector_agent_->addContext(GetContextLocal(), context_name);
 
   if (args_.flags.enableInspector) {
     TRACEV8RUNTIME_VERBOSE("Inspector enabled");
@@ -559,8 +551,8 @@ V8Runtime::~V8Runtime() {
 #if defined(_WIN32) && defined(V8JSI_ENABLE_INSPECTOR)
   {
     if (inspector_agent_) {
-      _ISOLATE_CONTEXT_ENTER
-      inspector_agent_->removeContext(context_.Get(isolate_));
+      IsolateLocker isolate_locker(this);
+      inspector_agent_->removeContext(GetContextLocal());
     }
   }
 #endif
@@ -601,7 +593,7 @@ V8Runtime::~V8Runtime() {
 }
 
 v8::Local<v8::String> V8Runtime::loadJavaScript(const std::shared_ptr<const jsi::Buffer> &buffer, std::uint64_t &hash) {
-  v8::EscapableHandleScope handle_scope(v8::Isolate::GetCurrent());
+  v8::EscapableHandleScope handle_scope(GetIsolate());
 
   bool isAscii = murmurhash(buffer->data(), buffer->size(), hash);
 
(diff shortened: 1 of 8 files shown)


src/V8JsiRuntime_impl.h
```diff
@@ -158,10 +158,14 @@ class V8Runtime : public facebook::jsi::Runtime {
 #endif
 
  public: // Used by NAPI implementation
-  v8::Global<v8::Context> &GetContext() {
+  inline v8::Global<v8::Context> &GetContext() {
```
NikoAri, line 161: >inline [](http://example.com/codeflow?start=2&length=6)

(nit) Why declaring explicit "inline"? Won't compiler inline as needed anyways?

NikoAri (approved):
:shipit:

src/V8JsiRuntime_impl.h
```diff
@@ -640,11 +644,38 @@ class V8Runtime : public facebook::jsi::Runtime {
   static void GCEpilogueCallback(v8::Isolate *isolate, v8::GCType type, v8::GCCallbackFlags flags);
 
  private:
+ // RAII wrappers for multi threaded support - order of the various scopes matters
+  struct IsolateScope {
+    IsolateScope(const V8Runtime *runtime) :
```
mganandraj, line 649: > * [](http://example.com/codeflow?start=32&length=2)

nit: This and IsolateLocker can take "const V8Runtime &"

src/V8JsiRuntime.cpp
```diff
@@ -1308,12 +1270,12 @@ std::string getFunctionName(v8::Isolate *isolate, v8::Local<v8::Function> func)
 
 jsi::Value
 V8Runtime::call(const jsi::Function &jsiFunc, const jsi::Value &jsThis, const jsi::Value *args, size_t count) {
-  _ISOLATE_CONTEXT_ENTER
+  IsolateLocker isolate_locker(this);
```
mganandraj, line 1273: >IsolateLocker isolate_locker(this); [](http://example.com/codeflow?start=2&length=35)

Do we use native promise in any scenarios ? How is the locker kept alive across multiple asynchronous micro tasks ?

tudorms: We don't use native promises in ReactNative (yet). Not sure what you mean about keeping the locker alive; we don't do that, we re-create the RAII wrapper on every entry into call (like we did for the HandleScope, Isolate Scope and Context Scope previously.

src/V8JsiRuntime_impl.h
tudorms: But all the callsites would need to pass `*this` which is ugly :)

mganandraj:
>   isolate_->Enter();

In Chakra, we used to enter the VM on each methods in runtime and exit at the end of method .. With V8, it looks like the isolate stays bound to a single thread.. but we take a lock on each entry to a method, which i think ensures that the isolate is running something only in one thread ..

In V8, does it make sense to exit and enter the isolate on each entry point to the runtime ?

I hope other threads such as JIT/GC etc. are also respecting the locks .. 

---
Refers to: src/V8JsiRuntime.cpp:411 in 26bc84a. [](commit_id = 26bc84a567bcf3d005c59dd15bcb6d4c4c333b37, deletion_comment = False)

src/V8JsiRuntime_impl.h
tudorms: I think you're right, it has no effect since it's a member function. But I wanted to be explicit, especially since this is a header.

mganandraj:
>   v8::Isolate *isolate = v8::Isolate::GetCurrent();

These methods can be called back during the remote-thread entry ? Does v8::Isolate::GetCurrent() return the correct isolate in that case ? 

---
Refers to: src/V8JsiRuntime.cpp:69 in 26bc84a. [](commit_id = 26bc84a567bcf3d005c59dd15bcb6d4c4c333b37, deletion_comment = False)

src/V8JsiRuntime_impl.h
mganandraj: :smile:

tudorms:
>   isolate_->Enter();

We do Enter and Exit on each entry point (Isolate::Scope is just a RAII that does Enter / Exit). This is needed here for simple cases where you create the Isolate and immediately use it for something (for example a few unit tests will fail if we remove it).

---
In reply to: [1273741990](https://github.com/microsoft/v8-jsi/pull/147#issuecomment-1273741990) [](http://example.com/codeflow?ancestors=1273741990,1273741990)

---
Refers to: src/V8JsiRuntime.cpp:411 in 26bc84a. [](commit_id = 26bc84a567bcf3d005c59dd15bcb6d4c4c333b37, deletion_comment = False)

tudorms:
>   v8::Isolate *isolate = v8::Isolate::GetCurrent();

I'm not sure this code is ever used. Tried setting enableMessageTracing = true but never hit this in the debugger.

---
In reply to: [1273743613](https://github.com/microsoft/v8-jsi/pull/147#issuecomment-1273743613) [](http://example.com/codeflow?ancestors=1273743613,1273743613)

---
Refers to: src/V8JsiRuntime.cpp:69 in 26bc84a. [](commit_id = 26bc84a567bcf3d005c59dd15bcb6d4c4c333b37, deletion_comment = False)

vmoroz:
It is a relatively big change. It would be nice to have an overview in the description of the PR.

src/V8JsiRuntime_impl.h
```diff
@@ -640,11 +644,38 @@ class V8Runtime : public facebook::jsi::Runtime {
...
+    v8::Context::Scope _context_scope;
+  };
+
+  struct IsolateLocker {
+    IsolateLocker(const V8Runtime *runtime) {
+      if (runtime->args_.flags.enableMultiThread) {
+        _locker = std::make_unique<v8::Locker>(runtime->GetIsolate());
+      }
+
+      _isolateScope = std::make_unique<IsolateScope>(runtime);
```
vmoroz, line 667: >make_unique [](http://example.com/codeflow?start=27&length=11)

Why do we to do 2 memory allocation each time we use the IsolateLocker? Is it possible to use field initializers like in the IsolateScope.

vmoroz:
Do we need a similar change for Node-API?

vmoroz (approved):
:shipit:

tudorms:
Added it now (a bit ugly due to the opaque napi_env_ thing, let me know if there's a cleaner approach).

---
In reply to: [1273855892](https://github.com/microsoft/v8-jsi/pull/147#issuecomment-1273855892) [](http://example.com/codeflow?ancestors=1273855892,1273855892)

src/V8JsiRuntime_impl.h
vmoroz: Could we use the std::optional for the conditional first case?

mganandraj: Note that Hermes has uses templated decorators ([https://github.com/facebook/hermes/blob/main/API/jsi/jsi/threadsafe.h](https://github.com/facebook/hermes/blob/main/API/jsi/jsi/threadsafe.h)) to take locks .. something like this might might help in avoiding locking costs for single threaded scenarios .. 

But it looks like not used in react-native right now ..

RN code explicitly takes locks before executing any JS code, so ensuring only one thread is running at the same time: 
[https://github.com/facebook/react-native/blob/main/ReactCommon/runtimeexecutor/ReactCommon/RuntimeExecutor.h](https://github.com/facebook/react-native/blob/main/ReactCommon/runtimeexecutor/ReactCommon/RuntimeExecutor.h)

Do V8 require taking IsolateLocks if the client code guarantees that only one thread is running JS at a time ?

mganandraj: > Note that Hermes has uses templated decorators (https://github.com/facebook/hermes/blob/main/API/jsi/jsi/threadsafe.h) to take locks .. something like this might might help in avoiding locking costs for single threaded scenarios ..
> 
> But it looks like not used in react-native right now ..
> 
> RN code explicitly takes locks before executing any JS code, so ensuring only one thread is running at the same time: https://github.com/facebook/react-native/blob/main/ReactCommon/runtimeexecutor/ReactCommon/RuntimeExecutor.h
> 
> Do we require taking IsolateLocks if the client code guarantees that only one thread is running JS at a time ?

> I don't intent to block this PR .. Just a question :)

mganandraj (approved):

src/V8JsiRuntime_impl.h
tudorms: v8::Locker is not copy constructible nor move constructible, so not sure how I'd optionally initialize a std::optional<v8::Locker>. I want to get rid of the dynamic allocs too; any ideas?

tudorms: If the Locker was ever used, it must always be used on that Isolate (so it doesn't matter that the client code "guarantees" it - it could only be a per-Isolate option).

PR #153, opened 2022-12-01
[0.65] Fix task runner for Node-API

vmoroz:
Currently the Node-API foreground task runner lifetime is depending on the `napi_env` lifetime.
This is wrong because some tasks must be run after the `napi_env` is destroyed.
This issue causes a crash in React Native for Windows when we do direct debugging.

In this PR we are fixing this issue by implementing a new task runner:
- Client code must call `v8_create_task_runner` function to create a new task runner and then pass it to `napi_ext_env_settings` when creating new `napi_env`. The code that internally uses `v8runtime::JSITaskRunner` is responsible for deleting it.
- The `v8_create_task_runner` captures external task scheduler and functions to run tasks and to destroy the external task scheduler.
- This is [a PR in RNW](https://github.com/microsoft/react-native-windows/pull/10966) that implements the new API.

An additional change in this PR: removed `NAPI_EXTERN` from all implementations as it may cause issues if the function signature is not the same in the .h and .cpp files. This issue was observed while working on the new `v8_create_task_runner` function.

Note that the new code has prefix `v8_` instead of previously used `napi_ext_`. The reason is that these functions are V8 JS Engine specific and will be never accepted as a part of the Node-API in Node.JS. Thus, the proposal is to use the `v8_` prefix instead. Ideally, we should replace `js_native_ext_api.h` with a new `v8_api.h` where all declarations will have the `v8_` prefix.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/153)

### DIFF
--- a/src/public/js_native_ext_api.h
+++ b/src/public/js_native_ext_api.h
@@ -25,22 +25,27 @@ typedef enum {
 
 typedef struct napi_ext_env_scope__ *napi_ext_env_scope;
 typedef struct napi_ext_ref__ *napi_ext_ref;
+typedef struct v8_task_runner_s *v8_task_runner_t;
+
+// A callback to run task
+typedef void(__cdecl *v8_task_run_cb)(void *task_data);
+
+// A callback to release task
+typedef void(__cdecl *v8_task_release_cb)(void *task_data);
+
+// A callback to post task to the task runner
+typedef void(__cdecl *v8_task_runner_post_task_cb)(
+    void *task_runner_data,
+    void *task_data,
+    v8_task_run_cb task_run_cb,
+    v8_task_release_cb task_release_cb);
+
+// A callback to release task runner
+typedef void(__cdecl *v8_task_runner_release_cb)(void *task_runner_data);
 
 // A callback to return buffer synchronously
 typedef void(
     __cdecl *napi_ext_buffer_callback)(napi_env env, uint8_t const *buffer, size_t buffer_length, void *buffer_hint);
-
-// A callback to run task
-typedef void(__cdecl *napi_ext_task_callback)(napi_env env, void *task_data);
-
-// A callback to schedule a task
-typedef void(__cdecl *napi_ext_schedule_task_callback)(
-    napi_env env,
-    napi_ext_task_callback task_cb,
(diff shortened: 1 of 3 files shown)


src/public/js_native_ext_api.h
```diff
@@ -92,8 +97,8 @@ typedef struct napi_ext_env_settings {
   // Size of this struct to allow extending it in future.
   size_t this_size;
 
-  // Custom scheduler of the foreground JavaScript tasks.
-  napi_ext_schedule_task_callback foreground_scheduler;
+  // Task runner for the foreground JavaScript tasks.
+  v8_task_runner_t foreground_task_runner;
```
mganandraj, line 101: >foreground_task_runner [](http://example.com/codeflow?start=19&length=22)

I'm curious to see the RNW changes which creates this .. Specifically will this take a reference to the runtime holder or the JS queue ?

mganandraj (approved):
:shipit:

src/public/js_native_ext_api.h
vmoroz: This is the RNW PR: https://github.com/microsoft/react-native-windows/pull/10966
Yes, it holds a reference to the JS queue.

tudorms (approved):

src/public/js_native_ext_api.h
```diff
@@ -25,22 +25,27 @@ typedef enum {
...
 
-// A callback to schedule a task
-typedef void(__cdecl *napi_ext_schedule_task_callback)(
-    napi_env env,
-    napi_ext_task_callback task_cb,
+// A callback to release task
+typedef void(__cdecl *v8_task_release_cb)(void *task_data);
+
+// A callback to post task to the task runner
+typedef void(__cdecl *v8_task_runner_post_task_cb)(
```
aeulitz, line 37: >v8_task_runner_post_task_cb [](http://example.com/codeflow?start=22&length=27)

OOC, why no equivalent to the 'delay_in_ms' and 'finalize_hint' parameters?

PR #161, opened 2022-12-29
Bump android jsi to 0.69-stable headers

rasaha91:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/161)

### DIFF
--- a/android/V8Runtime_shared.cpp
+++ b/android/V8Runtime_shared.cpp
@@ -417,6 +417,10 @@ namespace facebook { namespace v8runtime {
     return createPropNameID(stringRef(str));
   }
 
+  jsi::PropNameID V8Runtime::createPropNameIDFromSymbol(const jsi::Symbol& sym) {
+    throw jsi::JSINativeException("V8Runtime::createPropNameIDFromSymbol is not implemented!");
+  }
+
   std::string V8Runtime::utf8(const jsi::PropNameID& sym) {
     _ISOLATE_CONTEXT_ENTER
     return JSStringToSTLString(GetIsolate(), v8::Local<v8::String>::Cast(valueRef(sym)));
--- a/android/jsi/jsi.h
+++ b/android/jsi/jsi.h
@@ -274,6 +274,7 @@ class JSI_EXPORT Runtime {
       const uint8_t* utf8,
       size_t length) = 0;
   virtual PropNameID createPropNameIDFromString(const String& str) = 0;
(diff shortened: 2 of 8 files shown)


android/jsi/jsi.h
```diff
@@ -984,7 +992,7 @@ class JSI_EXPORT Value {
 
   /// Copies a Symbol lvalue into a new JS value.
   Value(Runtime& runtime, const Symbol& sym) : Value(SymbolKind) {
-    new (&data_.pointer) String(runtime.cloneSymbol(sym.ptr_));
+    new (&data_.pointer) Symbol(runtime.cloneSymbol(sym.ptr_));
```
dannyvv, line 995: Could this be a breaking change?

dannyvv (approved):

android/V8Runtime_shared.cpp
```diff
@@ -417,6 +417,10 @@ namespace facebook { namespace v8runtime {
     return createPropNameID(stringRef(str));
   }
 
+  jsi::PropNameID V8Runtime::createPropNameIDFromSymbol(const jsi::Symbol& sym) {
```
dannyvv, line 420: Should we implement this? This change does for JSI bindings

rasaha91: Hm, I didn't implement it because most of the symbol related APIs throw an unimplemented exception...

android/jsi/jsi.h
rasaha91: Honestly, I'm not sure. Why do you think it is breaking? @mganandraj in case you have thoughts

mganandraj: Ooh .. Looks like this was a typo in jsi header until 0.68 which got fixed in 0.69 .. The symbol was getting converted to String !

[https://github.com/facebook/react-native/commit/da48d0dc63de085f9fcc24774a063b36c86df65a](https://github.com/facebook/react-native/commit/da48d0dc63de085f9fcc24774a063b36c86df65a)

android/V8Runtime_shared.cpp
mganandraj: Our android implementation still don't support symbols .. @ranesh, i think you should file a task to track the limitation.. Hopefully we will switch to Hermes soon!!

mganandraj (approved):
:shipit:

PR #165, opened 2023-01-03
Use 0.70 jsi headers

rasaha91:
Bump JSI headers to 0.70-stable and add big int support

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/165)

### DIFF
--- a/android/V8Runtime_impl.h
+++ b/android/V8Runtime_impl.h
@@ -282,6 +282,19 @@ namespace facebook { namespace v8runtime {
       friend class V8Runtime;
     };
 
+    class V8BigIntValue final : public PointerValue {
+      V8BigIntValue(v8::Local<v8::BigInt> obj);
+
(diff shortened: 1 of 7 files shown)


mganandraj (approved):

tudorms:
There is duplicated effort between the \android folder and \src ; at some point, we had the main \src project build for Android too, might be worth revisiting that.

rasaha91:
> There is duplicated effort between the \android folder and \src ; at some point, we had the main \src project build for Android too, might be worth revisiting that.

Agreed, not sure why its so segregated at the moment. I'm also not sure what our plans are for supporting android v8 jsi once we switch over to hermes... we may abandon it. @mganandraj  in case you know.

PR #166, opened 2023-01-04
Upgrade android jsi headers to 0.71-stable

rasaha91:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/v8-jsi/pull/166)

### DIFF
--- a/android/V8Runtime_shared.cpp
+++ b/android/V8Runtime_shared.cpp
@@ -843,8 +843,10 @@ namespace facebook { namespace v8runtime {
     return objectRef(obj)->InternalFieldCount() == 1;
   }
 
+  constexpr auto c_nativeStateIndex = 0;
+
   std::shared_ptr<jsi::NativeState> V8Runtime::getNativeState(const jsi::Object& obj) {
-    std::shared_ptr<jsi::NativeState>* holder = static_cast<std::shared_ptr<jsi::NativeState>*>(objectRef(obj)->GetAlignedPointerFromInternalField(0));
+    std::shared_ptr<jsi::NativeState>* holder = static_cast<std::shared_ptr<jsi::NativeState>*>(objectRef(obj)->GetAlignedPointerFromInternalField(c_nativeStateIndex));
 
     return *holder;
   }
@@ -852,8 +854,8 @@ namespace facebook { namespace v8runtime {
   void V8Runtime::setNativeState(
       const jsi::Object& obj,
       std::shared_ptr<jsi::NativeState> nativeState) {
-    std::unique_ptr<std::shared_ptr<jsi::NativeState>> holder = std::make_unique<std::shared_ptr<jsi::NativeState>>(nativeState);
-    objectRef(obj)->SetAlignedPointerInInternalField(0, holder.get());
+    auto holder = std::make_unique<std::shared_ptr<jsi::NativeState>>(nativeState);
+    objectRef(obj)->SetAlignedPointerInInternalField(c_nativeStateIndex, holder.get());
   }
 
   jsi::ArrayBuffer V8Runtime::createArrayBuffer(std::shared_ptr<jsi::MutableBuffer>) {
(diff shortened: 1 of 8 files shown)


android/V8Runtime_shared.cpp
```diff
@@ -812,6 +812,54 @@ namespace facebook { namespace v8runtime {
...
+    bigIntRef(val)->Uint64Value(&lossless);
+    return lossless;
+  }
+
+  uint64_t V8Runtime::truncate(const jsi::BigInt& val) {
+    return bigIntRef(val)->Uint64Value(nullptr);
+  }
+
+  jsi::String V8Runtime::bigintToString(const jsi::BigInt&, int) {
+    throw jsi::JSINativeException("V8Runtime::bigintToString is not implemented!");
```
mganandraj, line 839: >throw jsi::JSINativeException("V8Runtime::bigintToString is not implemented!"); [](http://example.com/codeflow?start=4&length=79)

As v8::BigInt is an v8::Object, we can probably call ObjectProtoToString .. Worth a try !

android/V8Runtime_shared.cpp
```diff
@@ -812,6 +812,54 @@ namespace facebook { namespace v8runtime {
...
+  }
+
+  bool V8Runtime::hasNativeState(const jsi::Object& obj) {
+    return objectRef(obj)->InternalFieldCount() == 1;
+  }
+
+  std::shared_ptr<jsi::NativeState> V8Runtime::getNativeState(const jsi::Object& obj) {
+    std::shared_ptr<jsi::NativeState>* holder = static_cast<std::shared_ptr<jsi::NativeState>*>(objectRef(obj)->GetAlignedPointerFromInternalField(0));
```
mganandraj, line 847: >0 [](http://example.com/codeflow?start=147&length=1)

It would be neater to define a constant/MACRO ..

android/V8Runtime_shared.cpp
rasaha91: I just copied what the non-android impl was doing

android/V8Runtime_shared.cpp
rasaha91: I just copied the non-android impl. No idea what that zero represents.

mganandraj, line 847: >GetAlignedPointerFromInternalField(0 [](http://example.com/codeflow?start=112&length=36)

Is this guaranteed to return 0/NULL if native state is not set ?

android/V8Runtime_shared.cpp
rasaha91: I could do it, but I don't know how we would test it... what is the testing process for android?

android/V8Runtime_shared.cpp
```diff
@@ -812,6 +812,54 @@ namespace facebook { namespace v8runtime {
...
+
+    return *holder;
+  }
+
+  void V8Runtime::setNativeState(
+      const jsi::Object& obj,
+      std::shared_ptr<jsi::NativeState> nativeState) {
+    std::unique_ptr<std::shared_ptr<jsi::NativeState>> holder = std::make_unique<std::shared_ptr<jsi::NativeState>>(nativeState);
```
mganandraj, line 855: >std::unique_ptr<std::shared_ptr<jsi::NativeState>> [](http://example.com/codeflow?start=4&length=50)

nit: use auto

mganandraj (approved):
:shipit:

android/V8Runtime_shared.cpp
rasaha91: Will leave it for future consideration.

android/V8Runtime_shared.cpp
mganandraj: Essentially, every v8 objects provide a small number of data slots (0 to some n) for the host/embedder to store private data ..  

It is ok to leave as is ..

rasaha91: https://v8docs.nodesource.com/node-0.12/db/d85/classv8_1_1_object.html#a580ea84afb26c005d6762eeb9e3c308f sounds like if it isn't set, it is undefined behavior. I don't know how we would know if it was set.

rasaha91: we will definitely need to revisit the 0.71 implementation... for now I'm just getting it compiling to block the rn bump.

PR #178, opened 2023-06-20
Update to V8 11.5

tudorms:
We're pretty far behind on V8 versions, this brings us up to the latest upstream release. Tested via jsitests and in Office via unit tests.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,4 @@
 {
-    "version": "0.71.5",
-    "v8ref": "refs/branch-heads/10.9"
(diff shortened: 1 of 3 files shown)


vmoroz:
Just curious: how do you create the diff files?

vmoroz (approved):
:shipit:

tudorms:
> Just curious: how do you create the diff files?

I apply whatever can be applied with --reject, then fix everything else manually.

PR #189, opened 2024-02-09
[beta] Fixes for 12.2

tudorms:
Populate the beta branch with the Beta version of V8 (12.2). This version included additional checks for isolate entrance/exit consistency which we were failing before, because apparently we kept the Isolated "entered" all the time on the thread.

This change implies that we may create more than one Isolate (even on the same thread) so that's something to look out for (potential perf impact).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/v8-jsi/pull/189)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,5 +1,5 @@
 {
-    "version":  "0.71.13",
-    "v8ref":  "refs/branch-heads/12.1",
-    "buildNumber":  "285"
+    "version":  "0.0.1",
+    "v8ref":  "refs/branch-heads/12.2",
+    "buildNumber":  "1"
 }
--- a/src/node-api-jsi/NodeApiJsiRuntime.cpp
+++ b/src/node-api-jsi/NodeApiJsiRuntime.cpp
@@ -1528,7 +1528,6 @@ jsi::Value NodeApiJsiRuntime::callAsConstructor(const jsi::Function &func, const
 }
 
 jsi::Runtime::ScopeState *NodeApiJsiRuntime::pushScope() {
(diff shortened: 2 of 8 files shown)


Khalef1 (approved):

config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version":  "0.71.13",
-    "v8ref":  "refs/branch-heads/12.1",
-    "buildNumber":  "285"
+    "version":  "0.0.1",
```
vmoroz, line 2: Is it intentional that the version is changed to 0.0.1?

src/node-api-jsi/NodeApiJsiRuntime.cpp
```diff
@@ -1528,15 +1528,13 @@ jsi::Value NodeApiJsiRuntime::callAsConstructor(const jsi::Function &func, const
 }
 
 jsi::Runtime::ScopeState *NodeApiJsiRuntime::pushScope() {
-  NodeApiEnvScope scope{getEnv()};
```
vmoroz, line 1531: Do we pass all unit tests after this change?

tudorms: Yes. Seems like this was mismatched. enter1 -> enter2 -> exit1 -> exit2, and was crashing in V8 code.

config.json
tudorms: Yes, just in case this accidentally beta bits end up in the release pipeline, want to be able to spot it easily.

PR #208, opened 2024-08-20
[0.71] Fix Node-API based API shutdown sequence

vmoroz:
Fixed the shutdown sequence for the Node-API based API.
- Rollback all 12.6 related changes in the V8Runtime.
- Delete V8RuntimeEnv instance only after all its Node-API environments are destroyed.

Related changes:
- Some typo fixes in the README.md
- Restore the machine image pool for PR validation

### DIFF
--- a/.ado/windows-pr.yml
+++ b/.ado/windows-pr.yml
@@ -7,7 +7,8 @@ pr:
   - "*-stable"
 
 pool:
-  name: OEDevJeff-OfficePublic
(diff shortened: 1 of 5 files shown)


vmoroz:
Disabling the PR build verification. The image still uses VS 17.10 that cannot compile V8 code. We must have VS 17.11 for that. It shipped on 8/13/2024.

acoates-ms (approved):

PR #210, opened 2024-08-26
Remove all JSI and Node-API related changes

vmoroz:
The latest v8-jsi build 0.71.19 still causing issues in Office code.
Unfortunately, there is no way to repro the issues locally.
In this PR we rollback all the changes to JSI and Node-API that have been done since the latest v8-jsi release 0.71.14.
This is just to see that we can have a new v8-jsi build that passes Office tests.


vmoroz:
It seems that the 0.71.19 finally passed all verifications after re-run and it is good to go.
No need for this PR then. Closing.

